PR #2, opened 2019-04-29
Notice to install all dependencies in README

mboecker:
Adds a notice in the README that you should probably install all dependencies of the book.
This notice was previously only found on the first page of the book.


mllg:
Thanks!

mboecker:
Travis fails because of an error while installing mlr3ordinary, which occurs on my desktop computer too. (reported: https://github.com/mlr-org/mlr3ordinal/issues/3)

PR #3, opened 2019-04-29
work in progress: Adding a new learner chapter

mboecker:
work in progress


jakob-r:
I would devide the section into two parts
1.) Implement a new learner independently of the mlr3learners repo. Many people just want to draft something without having it ready to realease for real.
2.) Opening a PR with said learner in the mlr3learners repo. But maybe this even does not have to be covered in the book? @mllg

mllg:
Independently of the mlr3learners package: yes.

I not sure yet where such more technical documentation belongs. The outline of the book is not really clear to me yet. Maybe a vignette in mlr3learners is the better option for now.

jakob-r:
Regarding 1) I think having most information collected in one place is the better alternative. We can just use another chapter. Maybe 02 is "too" visible.
Regarding 2) That is so technical and too close to git etc. I would leave it out. Maybe in the README of this repo plus a link to the chapter in the book

mllg:
Creating a new chapter  "Extending mlr3" in the book seems a good fit.

mboecker:
> Creating a new chapter "Extending mlr3" in the book seems a good fit.

I think that this might be the best approach.

pat-s:
What's the status here? Does it need another review? Do you want to make more changes?

mboecker:
> What's the status here? Does it need another review? Do you want to make more changes?

We talked about moving the here added section to a new chapter because adding a new learner to mlr is not something every user would do.

pat-s:
Looks like we already have a page on that: https://mlr3book.mlr-org.com/learners-1.html

Probably @mllg 	wrote it?
It makes sense to continue on this base.

mboecker:
Yes that chapter improves this section, so I'll close this and the associated PR on mlr3 as well

PR #4, opened 2019-06-02
Fix weird R output and typo

mboecker:
this codeblock was producing weird output  because of the chunk options


pat-s:
Yes, it shouldn't be there, only in the setup chunk.

pat-s:
Thanks!

PR #11, opened 2019-06-06
Add a section in Introduction with a mlr->mlr3 lookup table

mboecker:
This should make it easier for long-time mlr users to switch to mlr3.
Is this roughly what you wanted, or should I go more in-detail?
I could also extend the table if you feel like this is missing crucial elements of mlr or mlr3.


bookdown/01-introduction.Rmd
```diff
@@ -22,3 +27,41 @@ Other _building blocks_ include:
 - Resampling methods
 
 These additional blocks are provided by our extension packages such as `r gh_pkg("mlr-org/mlr3tuning")`, `r gh_pkg("mlr-org/mlr3pipelines")` or `r gh_pkg("mlr-org/mlr3featuresel")`.ing.
+
+## Equivalent elements from `mlr`
```
pat-s: I would simply call it "transition table" or similar. 
Not everything will be equivalent in mlr3 and the list will also contain a lot of missing items.

bookdown/03-train-predict.Rmd
```diff
@@ -18,11 +18,11 @@ Additionally, this chapter includes the following use-cases
 The first step is to generate the following `mlr3` objects from the [task dictionary](#tasks) and the [learner dictionary](#learners), respectively:
 
 1. The classification task
-    ```{r}
+    ```{r train-predict-task}
```
pat-s: These changes were already merged in your prior PR right?

mboecker: Yes exactly I merely merged the branch into this one before the PR was merged

mboecker:
> The goal should be to have a table for every exported function in _mlr_.

Just to be clear, this means finding every function from https://mlr.mlr-org.com/reference/index.html in mlr3?

pat-s:
> Just to be clear, this means finding every function from mlr.mlr-org.com/reference/index.html in mlr3?

Somewhat yes. Not every function is mirrored and some old ones are now wrapped into one mlr3 class.

jakob-r:
I think the way it is programmed is a bit hard to read. First question, do we have/want A) a 1-1 relation, B) 1-n/n-1 or C) many to many in this table?

A) Do a named character vector `c("mlrfun" = "mlr3fun")` or vice versa. The table consists of `names(x)` and `x`.
B) Do a named list `list("mlrfun" = c("mlr3fun1", "mlr3fun2"))`. The table consists of `names(x)` and `lapply(x, paste, collapse = ", ")`. Of course you can make it more flexible here and make each list item a list with some description or whatever.
C) I havent made up my mind yet.

D) Do we even have to do this programmatically?

pat-s:
> D) Do we even have to do this programmatically?

Not really but is it so much easier not to do it programmatically?

Always happy to see improved layouts of this approach.

mboecker:
Whats the consensus here? I don't think the specific layout of the vectors is important because we dont expect users or anyone to use them without the book. For the book, I see the main usecase as searching (ctrl+F) for the item you want to have translated.

jakob-r:
> I don't think the specific layout of the vectors is important because we dont expect users or anyone to use them without the book.

As a user of the code I think it is very complicated to write the mapping because I would have to count where to put the new mlr3 function so it relates to the correct mlr function.

mboecker:
I agree. In this case, it would be my favorite option to go with n-m mapping using `c("a / b" = "c / d")`

pat-s:
I agree with both of your points here. 
I can see why the n-m mapping is actually easier for writing.
However, aren't we running into an ugly alignment with this?

jakob-r:
I don't get how you mean n-m mapping with `c("a / b" = "c / d")` but I am sure it will work.

I also don't now what the alignment issue should be :/

pat-s:
@mboecker 	Can you add one or two lines showing how you exactly would do it so that we can have a look at the visual output?

mboecker:
```r
t = c(
  "RegrTask / ClassifTask"         = "TaskRegr / TaskClassif",
  "makeRegrTask / makeClassifTask" = "TaskRegr$new() / TaskClassif$new()")
knitr::kable(names(t), t)
```
This is based on jakob's approach B. Output would look the same as before

jakob-r:
I think the table should be more `pure`. Becasue we also have `ClusterTask`, `SurvTask` and many more. For everything that is not in pure mlr3 we can add eg. `mlr3surv::`
If you want categories order the rows somehow or use something more complicated:
```r
t = "
category , mlr             , mlr3              , note
Task     , makeClassifTask , TaskClassif$new() , NA
Task     , RegrTask        , TaskRegr          , Class
Task     , makeRegrTaskm   , TaskRegr$new()    , NA
"
t = read.table(text =  t, sep = ",", header = TRUE)
t = knitr::kable(t)
kableExtra::collapse_rows(t, columns = 1)
```

pat-s:
2019-06-21-105113_screenshot

mllg:
Note: You can use `mlr3misc::rowwise_table()` to write down human-readable tables in R (idea stolen from `tibble::tribble`).

See https://github.com/mlr-org/mlr3/blob/master/R/mlr_reflections.R#L67 for an example.

mllg:
AFAICT `kableExtra` can only produce HTML tables. This is a road block if we want to compile the book into a PDF in the future. Is there an alternative package which can handle both output types?

jakob-r:
> AFAICT `kableExtra` can only produce HTML tables. 

It can produce both and nearly all stylings (or whatever you want to call them) translate into html and latex.

mllg:
Oh okay then.

pat-s:
Merging this now so that we have a first draft that we can continue on.

Future PRs should be of smaller size.

@mb706 	@mllg 	
If you implement new features in the future, it would be great if you could also update this list.
I know, feels like a 2nd NEWS file, but I can't track every new feature implementation to update this list alone. 
There is also still a lot missing and I hope @mboecker 	can add some more entries in the near future :)

PR #15, opened 2019-06-18
WIP: Add learner overview table (Appendix)

mboecker:
Add first version of the learner table.
I need feedback on these points:

- Display "Oob Error" as "OOB Error" ?
- Display predict type "Se" as something else?
- Still need to seperate these out by the package they come from (mlr3, learners, extralearners)


mllg:
Merged now, we can still cleanup later.

PR #20, opened 2019-07-09
Update pipeline docs

pfistfl:
Some more updates to pipeline section

- Disabled zoom for the html widget plots
- Fixed travis


bookdown/03-pipelines.Rmd
```diff
@@ -313,7 +321,7 @@ bagging = pred_set %>>%
 and plot again to see what happens:
 
 ```{r 03-pipelines-028, fig.width=7.5}
-bagging$plot(html = TRUE)
+bagging$plot(html = TRUE) %>% visInteraction(zoomView = FALSE) # disable zoom
```
pat-s: pipes on new line

bookdown/03-pipelines.Rmd
```diff
@@ -452,11 +460,12 @@ poi = PipeOpImpute$new(param_vals =
   list(method_num = "hist", method_fct = "sample", add_dummy = "all"))
 ```
 
-we can then add a learner to the Pipeop to obtain a learner equipped with automatic
-imputation.
+A learner can thus be equipped with automatic imputation of missing values
```
pat-s: This is an example sentence where the sentence is split into two lines? Most likely by sublime?

pfistfl: dont know. The replacement does not have this?

bookdown/03-pipelines.Rmd
pfistfl: done

bookdown/03-pipelines.Rmd
pat-s: What is "the replacement"? In the diff the new change breaks the sentence into two sentences, l. 456 and l.457

pfistfl: I think I got it now

bookdown/03-pipelines.Rmd
```diff
@@ -76,6 +74,15 @@ filter = mlr_pipeops$get("filter",
   param_vals = list(filter.frac = 0.5))
 ```
 
+## The Pipeline Operator: `%>>%` {#pipe-operator}
+
+Although it is possible to create intricate `Graphs` with edges going all over the place (as long as no loops are introduced), there is usually a clear direction of flow between "layers" in the `Graph`. It is therefore convenient to build up a `Graph` from layers, which can be done using the **`%>>%`** ("double-arrow") operator. It takes either a `PipeOp` or a `Graph` on each of its sides and connects all of the outputs of its left-hand side to one of the inputs each of its right-hand side--the number of inputs therefore must match the number of outputs.
```
pat-s: Please put every sentence on a new line :)

PR #26, opened 2019-07-15
cleanup fonts css

jakob-r:
only use 3 fonts
revert to default line hight

### DIFF
--- a/bookdown/style.css
+++ b/bookdown/style.css
@@ -1,18 +1,14 @@
-@import url('https://fonts.googleapis.com/css?family=Source Code Pro');
-@import url('https://fonts.googleapis.com/css?family=Montserrat');
-@import url('https://fonts.googleapis.com/css?family=Roboto');
-.book.font-family-0 {font-family: Montserrat, sans-serif;}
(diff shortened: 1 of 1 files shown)


pat-s:
You did not like the current custom line height and fonts? :/

mllg:
I agree with the font change, but the line height is probably more a question of taste?

pat-s:
ping @jakob-r

jakob-r:
1. Personally I prefer a bigger line hight and it is the default setting, so I changed that.
2. The fonts were strangely configured
2.1 no serifs for the serif style (family 0)
2.2 the list of replacement fonts was quite long e.g.: `Roboto, "Helvetica Neue", Helvetica, Arial, sans`

mllg:
We can do a poll which line height is best, but we should not change fonts based on personal preference. Deviations from the standard are usually not good.

berndbischl:
Michel is completely right. Please don't use an overcomplex setup now, or a nonstandard one. 
We should rather focus on content now, not font size optimization.

pat-s:
I did this several weeks ago and it did not take long. 
Also I do not think that this was an overcomplex change/setup :)
But fair enough now :duck:

PR #30, opened 2019-07-18
quickstart example

larskotthoff:
Very quick intro example on first page so that people see how to get started quickly.

I'm confused about the installation instructions. Is it really necessary to install the mlr3book package to run the examples? If so, why do the instructions say to load the mlr3 package after installing the mlr3book package?


mllg:
The book depends or suggests on packages which are not installed by mlr3, e.g., additional learners or packages with data sets.

I agree that this is a bit confusing in the introduction and should be re-worded, but I believe that installing the mlr3book package is a nice and convenient way to get everything in a single go.

pat-s:
> Is it really necessary to install the mlr3book package to run the examples?

An alternative way that could be mentioned is to install `mlr3verse`: `pak::pkg_install("mlr-org/mlr3verse")`.

larskotthoff:
Ah ok, that makes sense. I've clarified that in the text.

bookdown/index.Rmd
```diff
@@ -41,15 +41,34 @@ requireNamespace("kableExtra")
...
+
+The code examples in this book use a few additional packages that are not installed by default if you install `r mlr_pkg("mlr3")`.
+To run the code here, install the `r gh_pkg("mlr-org/mlr3book")` package using the `r cran_pkg("remotes")` package:
 
 ```{r index-002, eval = FALSE}
 remotes::install_github("mlr-org/mlr3book", dependencies = TRUE)
 ```
 
-After the installation, load the `mlr3` package and you are good to go:
+You don't need to load `r mlr_pkg("mlr3book")` though, just load the `r mlr_pkg("mlr3")` package and you are good to go:
```
pat-s: See https://github.com/mlr-org/mlr3book/pull/31#discussion_r305358342

jakob-r:
> An alternative way that could be mentioned is to install mlr3verse: pak::pkg_install("mlr-org/mlr3verse"). (this installs most of the mlr3 framework packages)
But then I might miss some example datasets, right?

pat-s:
> But then I might miss some example datasets, right?

Not if they are stored in the mlr3 packages. Calling this has a different purpose in the end: Install the mlr3 package framework, not "get all examples in the book to run".

Some people might now want to rerun all the examples in the book.
I guess we should at least list both options and make the difference clear.

- `remotes::install_github("mlr-org/mlr3book", dependencies = TRUE)`
	- `Will download 116 CRAN packages (90.51 MB), cached: 4 (7.73 MB).`
- `remotes::install_github("mlr-org/mlr3verse")`
	- `Will download 64 CRAN packages (51.62 MB), cached: 4 (7.73 MB).`

(I got the second line with `pak:::proj_install_dev()` from the repo roots, respectively.)

mllg:
Oh, Jakob is right. Florian has put an example data set for the regression use case in the mlr3book package. We might want to find better place for the data.

pat-s:
> Oh, Jakob is right. Florian has put an example data set for the regression use case in the mlr3book package. We might want to find better place for the data.

Basic example datasets should fit into _mlr3_? Example datasets for special tasks should go into their respective packages?

I would also vote for not putting example data into the book.

mllg:
The house sales data set is ~500 kB. Maybe we can just bundle it and upload it to CRAN? Who wants to experience the joy of uploading a package? @pfistfl?

jakob-r:
> Oh, Jakob is right. Florian has put an example data set for the regression use case in the mlr3book package. We might want to find better place for the data.

I was actually referring to the `nycflights13` suggest in the `DESCRIPTION` file which is used in `04-technical.Rmd:242`

larskotthoff:
Sounds like we need to agree on a wider point -- what is the purpose of the `mlr3book` package? Do you need to install this only to compile the book yourself, do you need it to run (at least some of) the examples in the book, or for extra datasets?

I would prefer the book package being required to build the book only, i.e. all code examples run if you install the `mlr3verse`.

jakob-r:
IMO the answer is in the middle:
- no datasets in any mlr3* package
- packages that supply the data goes into suggests
- when such data is used always make sure to load it explicitly

Consequences:
- Our packages stay small
- Users _can_ install `mlr3book` including the suggests and have all data packages available
- If user does not install `mlr3book` he will get an error `package xxx missing` which is no biggie, because he/she will just install the missing package

larskotthoff:
Ok, so then the instructions stay as they are -- install `mlr3book` package to get all dependencies, but only load `mlr3`?

berndbischl:
Here's a very simple suggestion. How about we do not overcomplicate this? Maybe just maybe a quick start (!) example can be constructed which uses an example task that Michel already put in mlr3.

And if that really does not suffice, and I guess for later example it really might not suffice: put the extra data sets into the mlr3book package on github? Problem solved?

larskotthoff:
Sounds good to me.

jakob-r:
> put the extra data sets into the mlr3book package on github? Problem solved?
IMHO this is not such a good idea because then
1. user finds code block in book
2. user wants to execute code
3. error because of missing data
4. user has mlr3verse loaded, gets frustrated and leaves

The minority of people will realise that you need to install the book to run the examples.

berndbischl:
Well what's your counter suggestion? Also we can, for my approach, put a simple sentence in bold into the preamble.

And it's handled in other books exactly like this....

berndbischl:
I mean I read your post above, of course. If you are saying we should explicitly load the Titanic package if we use the Titanic dataset: totally agree.

But I guess we are discussing data sets here which are not on cran. 

Hmm.

Here are options. 
1. We put them into cran packages ourselves. I seriously don't want that. 
2. We put them in the book. My suggestion. 
3. We load them from openml. 

Rethinking about this, 3 is best?

berndbischl:
3. is explicit and uses the correct tool for the problem.
Right?

jakob-r:
> put a simple sentence in bold into the preamble.

nobody reads that on a website. I guess most people wont read this book in a linear fashion.

> 1. We put them into cran packages ourselves. I seriously don't want that.

By all the datasets that are available on cran it suprises me that we find it necessary to use another one. Which might speak for the excellent properties of said data set as an example data set? So it could go on CRAN

> 3. is explicit and uses the correct tool for the problem.

How do we do it? Download the farf? Use OpenMlr together wirh mlr3?  Nontheless I like this approach for the reasons you mentioned.

I feel like we are wasting a lot of characters on this discussion (and I already feel sorry for it)

berndbischl:
Openml package does not depend on mlr2....

berndbischl:
Yeah it's an annoying discussion but we need to solve it, with a simple system. I can live with a book package, I can live with openml.
And if users don't want to read the first page of a book that's their problem not ours

mllg:
> Openml package does not depend on mlr2....

Yes, it does. Not even imports, it depends on it.

I'd favor option 1: putting extra data sets into packages and upload them to CRAN. This is a clean solution and a one time effort.

berndbischl:
> I'd favor option 1: putting extra data sets into packages and upload them to CRAN. This is a clean solution and a one time effort.

no. why? that circumvents openml as a project? it was clearly constructed for that. 

> Yes, it does. Not even imports, it depends on it.

that looks like a clear mistake? why not simply fix that?

mllg:
> that looks like a clear mistake? why not simply fix that?

@giuseppec ?

berndbischl:
I talked to Giuseppe he will look into it

berndbischl:
I talked to Giuseppe he will look into it

pat-s:
Long discussion.

Here are my thoughts:

- Most people won't install the mlr3book just to run the examples. Doing this should imo be only for the purpose of building the book yourself. Almost no user wants to go through all examples in the book in a linear fashion.
- Most people won't read that you need to install the book to get certain example datasets. Ofc we can say "that's bad for you because you did not read" but in the end thats our users which might leave. 
- If there is a `library()` call in the code users want to run, they are usually smart enough to install the missing package (in case they have loaded _mlr3_ or _mlr3verse_)
- For datasets not on CRAN, import them via OpenML to at least have a working code for the user without having to search how to get the data (the latter is most annoying, not c/p a few lines of OpenML code)

Regarding @larskotthoff question what to write: I would recommend installing the _mlr3verse_ and loading it to run most functionality of the book while keeping a medium sized footprint of dependencies. I mean the _mlr3verse_ represents our ecosystem in the end, not the _mlr3book_ package.

giuseppec:
I am working on the openml stuff.  Just noticed that this will probably take some time as I have to fix some new other issues due to server changes... Sorry for the delay. Will do my best to finish this soon.

pat-s:
@larskotthoff 	Is this superseeded by #31? If not, could you update here so we can merge? We would like to move some files once all PRs are merged.

larskotthoff:
Done, should be good to go.

giuseppec:
Donr, I have moved mlr to "suggests". Do you also need an updated OpenML package on CRAN?

berndbischl:
> Do you also need an updated OpenML package on CRAN?

yes

pat-s:
No changes left after the merge. Is that correct?

Well anyways, merge and move on for now :)

PR #31, opened 2019-07-18
Basics 1

larskotthoff:


bookdown/01-mlr3-basics.Rmd
```diff
@@ -1,28 +1,31 @@
 # mlr3 Basics {#basics}
 
-This chapter will teach you the essential building blocks, R6 classes and operations of `r mlr_pkg("mlr3")`.
-These include creating supervised ML tasks like classification/regression, performing training/prediction and cross-validate/benchmark different models.
+This chapter will teach you the essential building blocks, R6 classes, and operations of `r mlr_pkg("mlr3")`.
+This includes creating supervised machine learning tasks like classification and regression, training models and getting prediction on new data, and evaluating and comparison of different models through cross-validation and benchmarking.
```
pat-s:
```suggestion
This includes creating supervised machine learning tasks like classification and regression, training models and getting prediction on new data, and evaluating and comparing different models through cross-validation and benchmarking.
```

bookdown/01-mlr3-basics.Rmd
```diff
@@ -1,28 +1,31 @@
...
 
-A typical (simple) machine-learning workflow looks like this:
+A typical machine learning workflow looks like this:
```
pat-s: Should we make a global decision how we want ML to be written? I.e. either

- machine-learning
- machine learning

bookdown/01-mlr3-basics.Rmd
```diff
@@ -1,28 +1,31 @@
...
 knitr::include_graphics("images/ml_abstraction.png")
 ```
 
-Data is usually split up into training and testing portions.
-Using a learning algorithm we induce a model on training data, label the test instances with our model, and compare true test label to predicted labels.
-We obtain a scalar numeric score which quantifies the predictive power of our learner given the current data situation.
+The data, which `r mlr_pkg("mlr3")` encapsulates in tasks, is split into non-overlapping train and test sets to be able to evaluate models objectively &mdash; we are interested in models that generalize to new data rather than just memorizing the training data.
+The training data is given to a machine learning algorithm, called a learner in `r mlr_pkg("mlr3")`, that uses it to build a model of how the features of the data relate to the target values.
+This model is then used to produce predictions on the test data, which are compared to the ground truth values to assess the quality of the model.
+`r mlr_pkg("mlr3")` offers a number of different measures to quantify this quality; usually a numeric score.
```
pat-s: > usually a numeric score

If we mention it explicitly, should we also give an example? Which measures are not a numeric score?

bookdown/01-mlr3-basics.Rmd
```diff
@@ -1,28 +1,31 @@
...
+This process may be repeated several times, each time resampling different train and test sets from the original data set.
+Multiple resampling iterations allow to get a better generalization performance estimate for a particular type of model by quantifying its performance on different data.
```
pat-s:
```suggestion
Multiple resampling iterations allow to get a generalized performance estimate for a particular type of model by quantifying its performance on different data.
```

bookdown/01-mlr3-basics.Rmd
```diff
@@ -1,28 +1,31 @@
...
+* A [resampling](#resampling) specifies a series of train and test sets.
 
-* A [task](#tasks) wraps the data and stores additional information about it.
-* A [learners](#learners) interfaces R's many ML algorithms, allows train and predict operations on data and the setting and querying of hyperparameters.
-* A [measure](#measures) is a mapping from test labels and predicted labels on a test set to a numeric score.
-* A [resampling](#resampling) specifies through index sets the (repeated) splitting of the data into train and test sets.
-
-You will later learn how to enhance learners to full pipelines and tune hyperparameters, but for now we will stick to this simple workflow.
+In many cases, this simple workflow is not sufficient to deal with real-world data, which may require normalization, imputation of missing values, or feature selection.
+We will cover more complex workflows that allow to do this and much more later, but restrict ourselves to simple workflows like the one above here for the sake of clarity.
```
pat-s:
```suggestion
We will cover more complex workflows that allow to do this and even more later in the book.
For now, we restrict ourselves to simple workflows like the one above for the sake of clarity.
```

bookdown/index.Rmd
```diff
@@ -41,15 +41,34 @@ requireNamespace("kableExtra")
...
+
+The code examples in this book use a few additional packages that are not installed by default if you install `r mlr_pkg("mlr3")`.
+To run the code here, install the `r gh_pkg("mlr-org/mlr3book")` package using the `r cran_pkg("remotes")` package:
 
 ```{r index-002, eval = FALSE}
 remotes::install_github("mlr-org/mlr3book", dependencies = TRUE)
 ```
 
-After the installation, load the `mlr3` package and you are good to go:
+You don't need to load `r mlr_pkg("mlr3book")` though, just load the `r mlr_pkg("mlr3")` package and you are good to go:
```
pat-s: In fact all packages will be loaded in their respective chunks.
So this statement if only half true as is suggests that the user only needs to load _mlr3_.
I would write something like "all required packages will be loaded when required throughout the book".

bookdown/01-mlr3-basics.Rmd
berndbischl: in english you do not use hyphens for this.... so pls "machine learning" or even ML

bookdown/01-mlr3-basics.Rmd
larskotthoff: In particular I was thinking of confusion matrices. Yes, they're not really measures as such, but they fall into the same category. That said, I don't feel strongly about this if you want to change it.

bookdown/01-mlr3-basics.Rmd
larskotthoff: Don't agree with this change -- "generalized performance estimate" qualifies "performance estimate", but we're not doing anything to take an existing performance estimate and make it somehow more general. "Generalization performance estimate" is a different kind of performance estimate, i.e. a general one.

bookdown/index.Rmd
larskotthoff: I don't understand your point -- the sentence is saying that to run the examples in the book, only the mlr3 package needs to be loaded and not the mlr3book package. Is that not true? Does the user need to load any other packages?

That more packages may be loaded later on is a technical detail that the user doesn't need to know here.

bookdown/01-mlr3-basics.Rmd
berndbischl: i would keep the matrices and measures separate. both should be covered. but in terms of mlr3 a measure always returns a scalar 1-number score

bookdown/01-mlr3-basics.Rmd
berndbischl: lars is correct

bookdown/index.Rmd
mllg: We should make this clear:

* Install mlr3book if you (a) want to compile the book yourself or (b) do not want to be bugged while reading that some required packages are not installed.
* Install mlr3 if you just want to give it a shot and don't mind that you occasionally have to `install.packages()`.

Either way, load `mlr3`.

larskotthoff: So do you agree with the current wording?

mllg: Fine for me.

pat-s: Install _mlr3verse_ (?) if you just want to give it a shot and don't mind that you occasionally have to install.packages()

-> usually people need more than just _mlr3_ to get their work done

pat-s: @larskotthoff 	what's your view on this?
I am not pushing on a change if you like the current wording, this is just a suggestion to make it more clear imo and also include the _mlr3verse_ package at this point.

larskotthoff: Is there any advantage to installing mlr3verse instead of mlr3book? If we want to go this way, why have a mlr3book package in the first place?

pat-s: You only need to install half of the packages compared to mlr3book. See https://github.com/mlr-org/mlr3book/pull/30#issuecomment-514647772.

I do not see the point why we should force users to install the whole book dependencies.
Which user wants to run all the examples in the book? Users will always focus on specific sections.

And I would dislike seeing the _mlr3book_ being promoted as the "umbrella" package for the mlr3 pkg framework - this is not really its purpose.
We want to stay slim and lightweight and then suggest to install the mlr3book with all of its deps?
This does not make sense to me.
And yes, even if we only write such a "short" and "small" sentence in the README of the intro, the danger is high that this suggestion gets established and people install the whole dep chain.

The book package exists because 
1. Michel needed to write some wrappers for linking and building the book to work around some bookdown shortcomings
2. To have a quick way of installing all deps of the book (which are a lot)

larskotthoff: Ok, but then how do we handle this here? We should mention that to be able to run everything you should install the mlr3book package.

bookdown/index.Rmd
```diff
@@ -41,15 +41,34 @@ requireNamespace("kableExtra")
...
+learner$model
+
+predictions = learner$predict(task, row_ids = 121:150)
+predictions
+# accuracy of our model on the test set of the final 30 rows
+predictions$score("classif.acc")
+```
+
+The code examples in this book use a few additional packages that are not installed by default if you install `r mlr_pkg("mlr3")`.
+To run the code here, install the `r gh_pkg("mlr-org/mlr3book")` package using the `r cran_pkg("remotes")` package:
```
pat-s:
```suggestion
To run all examples in the book, install the `r gh_pkg("mlr-org/mlr3book")` package using the `r cran_pkg("remotes")` package:
```

bookdown/index.Rmd
pat-s:
```suggestion
Most often you do not want to run all code blocks of the book but just some parts of it.
In these cases it is easiest to install the _mlr3verse_ pkg (a wrapper package for the core packages of the _mlr3_ package framework) and post-install the required packages of the respective sections you want to run.

```r
remotes::install_github("mlr-org/mlr3verse")
```

```

pat-s: See my two comments/suggestions

pat-s:
Thanks to all :smile: :rocket:

PR #60, opened 2019-09-18
add information about learner_properties and predict_types

mboecker:
closes https://github.com/mlr-org/mlr3/issues/331


pat-s:
Thanks. Could you add the following information to `type = "response"`?

from `?predict.glm`
> the alternative "response" is on the scale of the response variable. 

which basically means that "response" in R denotes getting back the same type as the response variable is in (which is not always a numeric or a class label).

mboecker:
> Thanks. Could you add the following information to `type = "response"`?
> > the alternative "response" is on the scale of the response variable.

I feel like adding this information would confuse the user more than it would be doing good. When I request a prediction for the class labels, I'd be surprised if I got a response on any other scale.

pat-s:
With "response" you do not request class labels. You request simply the same scale that the response has.
If I understand your comment correctly, than your thinking of "response" might be wrong.

Anyway, the PR has been merged several days ago.
Thanks for your work.

PR #65, opened 2019-10-01
replace $get calls with respective sugar functions, closes #64

### LINKED ISSUE #64: Replace all `$get()` calls with their respective sugar functions
E.g. `mlr_learners$get()` -> `lrn()`

mboecker:
I'm unsure about the 30s quickstart example in index.Rmd, should we maybe use the dictionary style calls there?

### DIFF
--- a/bookdown/01-basics/01-resampling.Rmd
+++ b/bookdown/01-basics/01-resampling.Rmd
@@ -128,7 +128,7 @@ If you want to do that because you are coming from a specific modeling field, fi
 
 If your custom resampling method is widely used in your field, feel welcome to integrate it into one of the existing _mlr3_ extension packages or create your own extension package.
 
-A manual resampling instance can be created using the `"custom"` template from the `r ref("mlr_resamplings")` dictionary.
+A manual resampling instance can be created using the `"custom"` template.
 
 ```{r 01-resampling-012}
 resampling = rsmp("custom")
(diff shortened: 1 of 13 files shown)


jakob-r:
I would say sugar all the way. Only if we talk about dictionaries and how to use them, we should use the getters.

mboecker:
> I would say sugar all the way. Only if we talk about dictionaries and how to use them, we should use the getters.

ok thats what I thought :+1: I did replace them in the quickstart example

berndbischl:
> I would say sugar all the way. Only if we talk about dictionaries and how to use them, we should use the getters.

@jakob-r is correct. always use the sugar. but the book should have one dedicated section on the dictionaries and explain how they work and how the sugar calls relate to them. if that does not exist pls open an issue now. and write it

mboecker:
> Thanks. What jakob-r says is a good point - when introducing dictionaries we should briefly mention what happens in the background when one uses `lrn()` and friends, i.e. the ´$get()` calls. Could you add this to the section on dictionaries?

>but the book should have one dedicated section on the dictionaries and explain how they work and how the sugar calls relate to them. if that does not exist pls open an issue now. and write it

When dictionaries are introduced in the 01-tasks-chapter, we first mention the $get() function and then introduce the tsk() helper. Similarly, the conveniece functions lrn() and rsmp() are introduced. There is no dedicated chapter about the dictionaries as far as I can tell. Would you like this to go into the 04-technical chapter as its own section?

This can still me merged, I have opened [a new issue for this](https://github.com/mlr-org/mlr3book/issues/69).

berndbischl:
> Would you like this to go into the 04-technical chapter as its own section?

i think a dedicated mini chapter, in the technical part is certaintly warranted. we can also findish here, and open a clean issue and PR for that, i think that is best

PR #76, opened 2019-10-19
Enable cache

mllg:
Should be safe to use with `cache.lazy == FALSE`.
We might want to document how to clear the cache though.


pat-s:
On my Linux machine written changes are not updated in the preview, even when refreshing.

mllg:
This is a problem which has to do with the file layout (sub directories), not the cache. I'll revert the file layout changes we did some weeks ago.

PR #81, opened 2019-11-06
Readd pipelines

pfistfl:
This PR fixes the changes made to pipelines / mlr3 since it was updated last.
I have not incorporated the lates materials form `mlr-outreach`.
This can be done in subsequent steps.


bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
 # Pipelines {#pipelines}
 
-```{block, type='warning'}
-Work in progress :)
+```{r 03-pipelines-001_setup, message = FALSE, warning = FALSE, echo = FALSE, include = FALSE}
+knitr::opts_knit$set(root.dir = fs::dir_create(tempfile()))
+knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
+library(visNetwork)
+library(mlr3pipelines)
```
pat-s: I don't think we have any setup chunks for specific chapters anymore. If you need to load libraries, just do it user-visible in the place where its needed.

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+- Simultaneous path branching (data going both ways)
+- Alternative path branching (data going one specific way, controlled by hyperparameters)
+
+An extensive introduction to creating custom **PipeOps** (PO's) can be found in the [technical introduction](#extending-mlr3pipelines).
+
+Using methods from `r mlr_pkg("mlr3tuning")`, it is even possible to simultaneously optimize parameters of multiple processing units.
+
+A predecessor to this package is the [mlrCPO](https://github.com/mlr-org/mlrCPO) package, which works with mlr 2.x.
+Other packages that provide, to varying degree, some preprocessing functionality or machine learning domain specific language, are the
+`r cran_pkg("caret")` package and the related `r cran_pkg("recipies")`  project, and the `r cran_pkg("dplyr")` package.
```
pat-s: wrong linebreak

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+
+An example for a Pipeline that can be constructed using
+`r mlr_pkg("mlr3pipelines")` is depicted below:
```
pat-s: wrong linebreak

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+
+```{r 03-pipelines-011}
+graph = Graph$new()$
+  add_pipeop(mutate)$
+  add_pipeop(filter)$
+  add_edge("mutate", "variance")  # add connection mutate -> filter
+```
+
+
+The much quicker way is to use the `%>>%` operator to chain POs or `r ref("Graph")` s.
```
pat-s:
```suggestion
The much quicker way is to use the `%>>%` operator to chain POs or `r ref("Graph")`s.
```

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+The same result as above can be achieved by doing
+
+```{r 03-pipelines-012}
+graph = mutate %>>% filter
+```
+
+
```
pat-s: unneeded empty line

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+ps = ParamSet$new(list(
+  ParamDbl$new("classif.rpart.cp", lower = 0, upper = 0.05),
+  ParamDbl$new("variance.filter.frac", lower = 0.25, upper = 1)
+))
+```
+
+After having defined the `PerformanceEvaluator`, a random search with 10 iterations is created.
+For the inner resampling, we are simply doing holdout (single split into train/test) to keep the runtimes reasonable.
+
+
```
pat-s: unneeded empty line

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+library("mlr3tuning")
+instance = TuningInstance$new(
+  task = task,
+  learner = glrn,
+  resampling = rsmp("holdout"),
+  measures = msr("classif.ce"),
+  param_set = ps,
+  terminator = term("evals", n_evals = 20)
+)
+```
```
pat-s: add empty line below

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+(graph %>>% mlr_pipeops$get("unbranch", c("nop", "pca", "scale")))$plot(html = TRUE) %>%
+  visInteraction(zoomView = FALSE) # disable zoom
+```
+
+### Model Ensembles {#pipe-model-ensembles}
+
+We can leverage the different operations presented to connect POs to form powerful graphs.
+
+Before we go into details, we split the task into train
+and test indices.
```
pat-s: wrong linebreak

bookdown/03-pipelines/03-pipelines.Rmd
```diff
@@ -1,5 +1,580 @@
...
+```
+
+We can now copy this operation 10 times using `r ref("greplicate")` .
+
+```{r 03-pipelines-029}
+pred_set = greplicate(single_pred, 10L)
+```
+
+Afterwards we need to aggregate the 10 pipelines
+to form a single model:
```
pat-s: wrong linebreak

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
-## Extending mlr3pipelines {#extending-mlr3pipelines}
+1## Extending mlr3pipelines {#extending-mlr3pipelines}
```
pat-s:
```suggestion
## Extending mlr3pipelines {#extending-mlr3pipelines}
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+### General Case Example: `PipeOpCopy` {#ext-pipeopcopy}
+
+A very simple yet useful `PipeOp` is `PipeOpCopy`, which takes a single input and creates a variable number of output channels, all of which receive a copy of the input data.
+It is a simple example that showcases the important steps in defining a custom `PipeOp`.
+We will show a simplified version here, **`PipeOpCopyTwo`**, that creates exactly two copies of its input data.
+
+The following figure visualizes how our `PipeOp` is situated in the `Pipeline` and the significant in- and outputs.
+
+![Pipeop Copy2](images/po_multi_viz.png){ width=90% }
```
pat-s: Does this work in the mlr3book? Looks like some javascript magic (the width setting)

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+#### Train and Predict
+
+Both `train()` and `predict()` will receive a `list` as input and must give a `list` in return.
+According to our `input` and `output` definitions, we will always get a list with a single element as input, and will need to return a list with two elements. Because all we want to do is create two copies, we will just create the copies using `c(inputs, inputs)`.
+
+Two things to consider:
+
+- The `train()` function must always modify the `self$state` variable to something that is not `NULL` or `NO_OP`. This is because the `$state` slot is used as a signal that `PipeOp`  has been trained on data, even if the state itself is not important to the `PipeOp` (as in our case). Therefore, our `train()` will set `self$state = list()`.
+
+- It is not necessary to "clone" our input or make deep copies, because we don't modify the data. However, if we were changing a reference-passed object, for example by changing data in a `Task`, we would have to make a deep copy first. This is because a `PipeOp` may never modify its input object by reference.
```
pat-s: Would be great if we could also stick to "one sentence one line" here. Just indent the sentence and the listing should be fine.

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+In the second case, one can optionally also overload the `select_cols` function, which chooses which of the incoming `Task`'s features are given to the `train_dt` / `predict_dt` functions.
+
+The following will show two examples: `PipeOpDropNA`, which removes a `Task`'s rows with missing values during training (and implements `train_task` and `predict_task`), and `PipeOpScale`, which scales a `Task`'s numeric columns (and implements `train_dt`, `predict_dt`, and `select_cols`).
+
+#### Example: `PipeOpDropNA`
+
+Dropping rows with missing values may be important when training a model that can not handle them.
+
+Because `mlr3` `Task`s only contain a view to the underlying data, it is not necessary to modify data to remove rows with missing values. Instead, the rows can be removed using the `Task`'s `$filter` method, which modifies the `Task` in-place. This is done in the `train_task` function. We take care that we also set the `$state` slot to signal that the `PipeOp` was trained.
```
pat-s:
```suggestion
Because `mlr3` `Task`s only contain a view to the underlying data, it is not necessary to modify data to remove rows with missing values. 
Instead, the rows can be removed using the `Task`'s `$filter` method, which modifies the `Task` in-place. 
This is done in the `train_task` function. 
We take care that we also set the `$state` slot to signal that the `PipeOp` was trained.
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+The `predict_task` function does not need to do anything; removing missing values during prediction is not as useful, since learners that cannot handle them will just ignore the respective rows. Furthermore, `mlr3` expects a `Learner` to always return just as many predictions as it was given input rows, so a `PipeOp` that removes `Task` rows during training can not be used inside a `GraphLearner`.
```
pat-s:
```suggestion
The `predict_task` function does not need to do anything; removing missing values during prediction is not as useful, since learners that cannot handle them will just ignore the respective rows. 
Furthermore, `mlr3` expects a `Learner` to always return just as many predictions as it was given input rows, so a `PipeOp` that removes `Task` rows during training can not be used inside a `GraphLearner`.
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+When we inherit from `PipeOpTaskPreproc`, it sets the `input` and `output` `data.table`s for us to only accept a single `Task`. The only thing we do during `initialize()` is therefore to set an `id` (which can optionally be changed by the user).
```
pat-s:
```suggestion
When we inherit from `PipeOpTaskPreproc`, it sets the `input` and `output` `data.table`s for us to only accept a single `Task`. 
The only thing we do during `initialize()` is therefore to set an `id` (which can optionally be changed by the user).
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+The complete `PipeOpDropNA` can therefore be written as follows. Note that it inherits from `PipeOpTaskPreproc`, unlike the `PipeOpCopyTwo` example from above:
```
pat-s:
```suggestion
The complete `PipeOpDropNA` can therefore be written as follows. 
Note that it inherits from `PipeOpTaskPreproc`, unlike the `PipeOpCopyTwo` example from above:
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+We test this by feeding it to a new `Graph` that uses `PipeOpDropNA`.
+```{r 05-extending-023}
+gr = Graph$new()
+gr$add_pipeop(PipeOpDropNA$new())
+
+filtered_task = gr$train(sitask)[[1]]
+print(filtered_task$data())
+```
+
+
```
pat-s: unneeded empty line

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+#### Example: `PipeOpScaleAlways`
+
+An often-applied preprocessing step is to simply **center** and/or **scale** the data to mean $0$ and standard deviation $1$.
+This fits the `PipeOpTaskPreproc` pattern quite well.
+ Because it always replaces all columns that it operates on, and does not require any information about the task's target, it only needs to overload the `train_dt` and `predict_dt` functions.
+ This saves some boilerplate-code from getting the correct feature columns out of the task, and replacing them after modification.
+
+Because scaling only makes sense on numeric features, we want to instruct `PipeOpTaskPreproc` to give us only these numeric columns.
+We do this by overloading the `select_cols` function: It is called by the class to determine which columns to give to `train_dt` and `predict_dt`.
+Its input is the `Task` that is being transformed, and it should return a `character` vector of all features to work with. When it is not overloaded, it uses all columns; instead, we will set it to only give us numeric columns.
```
pat-s:
```suggestion
Its input is the `Task` that is being transformed, and it should return a `character` vector of all features to work with. 
When it is not overloaded, it uses all columns; instead, we will set it to only give us numeric columns.
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+Because the `levels()` of the data table given to `train_dt` and `predict_dt` may be different from the levels `task`'s levels, these functions must also take a `levels` argument that is a named list of column names indicating their levels. When working with numeric data, this argument can be ignored, but it should be used instead of `levels(dt[[column]])` for factorial or character columns.
```
pat-s:
```suggestion
Because the `levels()` of the data table given to `train_dt` and `predict_dt` may be different from the levels `task`'s levels, these functions must also take a `levels` argument that is a named list of column names indicating their levels. 
When working with numeric data, this argument can be ignored, but it should be used instead of `levels(dt[[column]])` for factorial or character columns.
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+Just like `PipeOpTaskPreproc`, `PipeOpTaskPreprocSimple` offers the possibility to instead overload the `get_state_dt(dt, levels)` and `transform_dt(dt, levels)` functions (and optionally, again, the `select_cols(task)` function) to operate on `data.table` feature data instead of the whole `Task`.
+
+Even some methods that do not use `PipeOpTaskPreprocSimple` *could* work in a similar way: The `PipeOpScaleAlways` example from above will be shown to also work with this paradigm.
+
+#### Example: `PipeOpDropConst`
+
+A typical example of a preprocessing operation that does almost the same operation during training and prediction is an operation that drops features depending on a criterion that is evaluated during training.
+One simple example of this is dropping constant features. Because the `mlr3` `Task` class offers a flexible view on underlying data, it is most efficient to drop columns from the task directly using its `$select()` function, so the `get_state_dt(dt, levels)` / `transform_dt(dt, levels)` functions will *not* get used; instead we overload the `get_state(task)` and `transform(task)` functions.
```
pat-s:
```suggestion
One simple example of this is dropping constant features. 
Because the `mlr3` `Task` class offers a flexible view on underlying data, it is most efficient to drop columns from the task directly using its `$select()` function, so the `get_state_dt(dt, levels)` / `transform_dt(dt, levels)` functions will *not* get used; instead we overload the `get_state(task)` and `transform(task)` functions.
```

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
```diff
@@ -1,5 +1,470 @@
...
+
+The `get_state()` function's result is saved to the `$state` slot, so we want to return something that is useful for dropping features.
+We choose to save the names of all the columns that have nonzero variance. For brevity, we use `length(unique(column)) > 1` to check whether more than one distinct value is present; a more sophisticated version could have a tolerance parameter for numeric values that are very close to each other.
```
pat-s:
```suggestion
We choose to save the names of all the columns that have nonzero variance. 
For brevity, we use `length(unique(column)) > 1` to check whether more than one distinct value is present; a more sophisticated version could have a tolerance parameter for numeric values that are very close to each other.
```

pfistfl:
Fixed, the `{width}` stuff seems to work.

bookdown/05-extending/05-extending-mlr3pipelines.Rmd
pfistfl: Seems to work

mllg:
Can we use `knitr::include_graphics()`? This works for markdown + latex, you can control the width/height and it automatically uses png for html output and pdf for pdf output (we might even patch it in the future to use svg instead of png for html).

mllg:
NB1: We need to write this down in the README.

NB2: Always store images also in a vector format like svg, even if you do not use them in vector format in the book. Otherwise, we cannot re-use or modify images in the future. I guess the images in this PR are even available as tikz, so it would be great to have a small "standalone" tikz image.

pfistfl:
Currently, all the figures come from google presentations, i.e. afaik no tikz or svg.
I guess we will have somebody replace the figures at a later stage.

pat-s:
@mllg done in https://github.com/mlr-org/mlr3book/commit/519cc19c7f3695c349501b2ee4831f22d11be488

PR #83, opened 2019-11-09
pipelines modifications.

danielsaggau:
Mostly:
- shortening sentences
- punctuation
- bullet points
- formating


mllg:
Thanks, merging this now, but I wonder why there are FIXMEs in the book. Is there an issue for this?

pat-s:
@pfistfl Could you have a look at the FIXME's in the text? They look like as some words are missing in important places..

PR #90, opened 2019-11-18
correct figure scaling for pdf and html

eddelbuettel:
The simple addition of `out.width="x%"` allows use of 'oversized' graphics which will get shrunk correctly in both html and pdf mode.


mllg:
Thanks!

pat-s:
Very neat. Haven't seen this one before!

PR #91, opened 2019-11-18
three more correct figure scaling for pdf and html

eddelbuettel:
Similar to #90 adding three more such adjustments.  Now the pdf has all figures nicely 'visible'.


larskotthoff:
Great, thanks!

PR #92, opened 2019-11-20
one small typo/grammar fix and one added word

eddelbuettel:
Simple case of an 's that needs to be a simple plural s, and an added word in same sentence.


larskotthoff:
Thanks!

PR #94, opened 2019-11-24
Replace mlr3survival with mlr3proba

adibender:
- some errors in appendix remain

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,7 +48,7 @@ Suggests:
     mlr3misc,
     mlr3ordinal,
     mlr3pipelines,
-    mlr3survival,
+    mlr3proba,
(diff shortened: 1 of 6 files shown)


pat-s:
Thanks. @RaphaelS1 Could you maybe help here resolving the issues?

RaphaelS1:
Sure I can see the problem. I can fix this soon

mllg:
Travis is now complaining that `pec` is not installed.

I'd suggest removing `"surv.rpart"` from the benchmark so that we do not need the `pec` dependency for the book.

RaphaelS1:
Changed to `surv.kaplan` as this is a sensible baseline and only requires `survival`

mllg:
Thanks everyone!

PR #97, opened 2019-11-27
"Code and description in conflict"

adibender:
The text explicitly states that default for `xval = 10L`, but that it should not be set in `ParamInt$new()`, while in the code reads 

```r
initialize = function(id = "classif.rpart") {
      ParamInt$new(id = "xval", default = 0L, lower = 0L, tags = "train")
```


mllg:
Thanks!

PR #113, opened 2020-01-12
Change geometry and font-size for PDF output

pat-s:


pat-s:
@mllg Related to our discussion of only changing the font-size for the code blocks and not the inline code: https://stackoverflow.com/questions/27000906/how-to-set-the-font-size-for-code-blocks-in-pandoc-markdown

Can't render a preview currently due to https://github.com/mlr-org/mlr3/issues/414 so if you want to give it a try - I think there is not more do to than to add the posted snipped to `preamble.tex`.

pat-s:
@mllg Tested locally, looks good to me

pat-s:
@mllg I am happy now with the font size and margins.

The `geometry` setting has only an effect in `index.Rmd`, not in `output.yml`.

- Main body font size: 9pt
- code block font: relative to it with size `\small`

pat-s:
Merging, feel free to comment/make adjustments afterwards.

bookdown/index.Rmd
```diff
@@ -15,6 +14,12 @@ cover-image: "block.png"
 favicon: "favicon.ico"
 apple-touch-icon: "apple-touch-icon.png"
 apple-touch-icon-size: 180
+# 'geometry' has only an effect when set here, not in output.yml
+geometry:
```
mllg: Why are we doing this via yaml? Wouldn't it be better to put stuff like this in the preamble?

bookdown/_output.yml
```diff
@@ -29,3 +29,5 @@ bookdown::pdf_book:
   latex_engine: xelatex
   highlight: haddock
   keep_tex: yes
+  documentclass: srcbook
+  fontsize: 9pt
```
mllg: Why are we doing this via yaml? Wouldn't it be better to put stuff like this in the preamble?

bookdown/index.Rmd
mllg: Also, I still don't get why using custom margins is better than defaulting to some presets koma script provides (DIV=14).

bookdown/preamble.tex
```diff
@@ -88,3 +86,5 @@
 \urlstyle{tt}
 
 \makeatother
+
+ \renewenvironment{Shaded} {\begin{snugshade}\small} {\end{snugshade}}
```
mllg: What is this about?

bookdown/index.Rmd
pat-s: > Why are we doing this via yaml? Wouldn't it be better to put stuff like this in the preamble?

Imo we should use the pandoc way of changing settings rather than using LaTeX code in `preamble.tex`. 

Also some settings are otherwise set twice (pandoc default + preamble.tex) and will cause an error.

> Also, I still don't get why using custom margins is better than defaulting to some presets koma script provides (DIV=14).

I don't know why the defaults should always be the best setting. Most often, they are just defaults...

bookdown/_output.yml
pat-s: Otherwise we have two `\documentclass()` class in the final .tex (pandoc default + custom setting).

bookdown/preamble.tex
pat-s: Code block resizing (`\small`). 

`snugshade` is the env for the grey background of code blocks.

bookdown/index.Rmd
mllg: > Imo we should use the pandoc way of changing settings rather than using LaTeX code in `preamble.tex`.

I don't see an advantage if we already have a `preamble.tex`. You can put everything (well, maybe not the documentclass) into the preamble, but you can only put a tiny fraction of latex into the yaml. We should keep settings  together.


> I don't know why the defaults should always be the best setting. Most often, they are just defaults...

... defaults carefully chosen by typographers. KOMA goes back to the work of Jan Tschichold. See his wiki page (https://en.wikipedia.org/wiki/Jan_Tschichold) for considerations on how to determine good margins.

pat-s: > We should keep settings together.

That's what I did following the YAML approach.
Feel free to move it.
But be aware that this can introduce problems for some settings due to conflicts with pandocs template settings.

If you dislike the new margins, feel free to remove them again. 
I will still dislike the typographers defaults ;)

PR #114, opened 2020-01-12
clean travis.yml

pat-s:
- `sudo` has no effect anymore
- from my experience we do not need a hardcoded pandoc version. Removing this makes us dynamic to the upstream updates


mllg:
Thanks

PR #120, opened 2020-01-19
PDF: Add titlepage

pat-s:
Move some settings to `output.yml` to make them output type specific.


bookdown/_output.yml
```diff
@@ -23,8 +26,12 @@ bookdown::gitbook:
        <li><a href="./">Made with <i class="fas fa-heart"></i> for open-source </a></li>
 
 bookdown::pdf_book:
+  author: ""
```
mllg: Why is this empty now?

bookdown/before_body.tex
```diff
@@ -0,0 +1,38 @@
+\begin{titlepage}
```
mllg: KOMA supports title pages. Why don't we use that?

bookdown/_output.yml
pat-s: Otherwise we get a single entry on the first page with just the author name. This is how pandoc does it.
I'd prefer to start the book with the titlepage.

bookdown/before_body.tex
pat-s: I did not know that and had the snippet from my MA thesis still available.

PR #124, opened 2020-01-31
Remove use cases

mllg:
... and more cleanups. Use cases will go into mlr3gallery.


mllg:
Merging although travis is failing. It seems to be a CI error where the package is not installed correctly prior to rendering the book.

PR #158, opened 2020-03-04
updated transition table #1

henrifnk:
Updated functions and implemented links.
Rendering pdf is still not possible


henrifnk:
outdated version.
If a tansition table should be requested to add to mlr3book a new PR should be created.

PR #168, opened 2020-04-06
mlr3pipelines greplicate --> ppl("greplicate")

sumny:
Should be merged when mlr-org/mlr3pipelines#396 is merged.


pfistfl:
@mllg I guess this can be merged

PR #177, opened 2020-07-24
Correct list of resampling strategies

### LINKED ISSUE #176: LOO CV resampling absent from mlr_resamplings
`bookdown/02-basics-resampling.Rmd` says:

> predefined resampling strategies: Cross-validation, Leave-one-out cross validation, Repeated cross-validation, Out-of-bag bootstrap and other variants (e.g. b632), Monte-Carlo cross-validation and Holdout.

, but `mlr_resamplings` does not seem to list Leave-one-out cross validation resampling for `mlr3`, at least not explicitly, as it is defined in the `mlr` package. I checked every strategy, suspecting that maybe it was hidden under another name, but none appears to match LOO for me. Is this an error in the book or am I missing something?

I can imagine LOO being a special case of a more general strategy (like an n-fold CV with the number of folds equal to the number of data points). If this is how it should be performed in `mlr3`, IMHO this needs to be given a separate note in the book.

mllg:
Fixes #176


mllg:
Requires CRAN release of mlr3 0.5.0 :worried:

PR #178, opened 2020-08-05
Update index, include density

mllg:


mllg:
Blocked by insufficient mlr3viz version on CRAN. Needs 0.2.0.

PR #184, opened 2020-09-02
Enable density chapter

mllg:


mllg:
Fixed directly in master.

PR #190, opened 2020-09-09
change refs from org to mlr3extralearners

RaphaelS1:
* This includes a big re-write of extending-learners that will need to be reviewed properly when built


RaphaelS1:
I've built and read through the updates to the extending learners section and am happy for it to be reviewed now

RaphaelS1:
Will update with latest Prediction methods once mlr3 0.6.0 on CRAN

PR #197, opened 2020-10-09
add jm comments

jannes-m:
Hey @pat-s ,
I have only slightly updated the text. Hope it helps.
Best,
Jannes

### DIFF
--- a/bookdown/07-special-spatiotemp.Rmd
+++ b/bookdown/07-special-spatiotemp.Rmd
@@ -11,8 +11,8 @@ In the `r cran_pkg("mlr3")` framework, the following packages contribute to this
 - `r gh_pkg("mlr-org/mlr3forecasting")` (time-series support)
 - `r gh_pkg("mlr-org/mlr3raster")` (enhanced spatial prediction)
 
-The following sections will introduce potential pitfalls of spatiotemporal data in machine learning and present possible workarounds.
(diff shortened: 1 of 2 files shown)


bookdown/07-special-spatiotemp.Rmd
```diff
@@ -84,6 +86,8 @@ rr_sp = resample(
 rr_sp$aggregate(measures = msr("classif.ce"))
 ```
 
+<!--JM: shouldn't we use `benchmark_grid()` to ensure that the same partitions are used for both resamplings?-->
```
jannes-m, line 89: Shouldn't we use `benchmark_grid()` to ensure that the same partitions are used for both resamplings?

pat-s: `benchmark_grid()` is just a wrapper around `resample()`. I think calling the same function two times with just a different method is more clear for demonstration purposes. 

Why do you expect the partitions to be equal here? We are using different resampling methods so the partitions will be different.

pat-s (approved):
Thanks Jannes! Did not expect that :) Seems you are watching closely :)

Reads way better now. My wording was sloppy.

I'll happily merge this and continue with Marcs comments.

bookdown/07-special-spatiotemp.Rmd
jannes-m: You are absolutely right, my bad!

PR #202, opened 2020-10-30
remove extralearner learners

RaphaelS1:
Removes any learners used in mlr3extralearners


pat-s:
@RaphaelS1 Have a look at the build - would be great if we could get this merged quickly :)

RaphaelS1:
The simsurv dependency isn't required, I can change the example to remove it.
I seem to remember this PR needs the latest proba CRAN version which will be out today as it also fixes the bugs that @be-marc pointed out yesterday. Proba will be on CRAN today, but then viz also needs updating....

pat-s:
> The simsurv dependency isn't required, I can change the example to remove it.

Yes please.

> Proba will be on CRAN today, but then viz also needs updating....

You mean mlr3viz needs a CRAN update? Then please tell @mllg :)

RaphaelS1:
proba on CRAN and viz update just merged by @mllg , so hopefully this can be merged tomorrow or day after depending on viz

PR #209, opened 2020-11-11
Ps shortcut doc

mb706:
documents `ps()` and `to_tune()`. Furthermore usees `ps()` instead of `ParamSet$new()` in all places in the book except the more technical chapters.


mb706:
(build fails because it needs github version of paradox)

[rendered html of searchspace chapter](https://github.com/mlr-org/mlr3book/files/5520946/searchspace.zip)

bookdown/03-optimization-hyperband.Rmd
```diff
@@ -183,26 +183,26 @@ But it is also possible to define a custom `r ref("Sampler")` for each hyperpara
...
   trm("none") # hyperband terminates itself
 )
 
 # beta distribution with alpha = 2 and beta = 5
 # categorical distribution with custom probabilities
 sampler = SamplerJointIndep$new(list(
-  Sampler1DRfun$new(params[[2]], function(n) rbeta(n, 2, 5)),
-  Sampler1DCateg$new(params[[3]], prob = c(0.2, 0.3, 0.5))
+  Sampler1DRfun$new(params$params$eta, function(n) rbeta(n, 2, 5)),
+  Sampler1DCateg$new(params$params$booster, prob = c(0.2, 0.3, 0.5))
```
jakob-r, line 205: Could you call `params` `ps` instead? it becomes confusing here.

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,223 @@
...
+
+An important distinction to keep in mind when defining `r ref("ParamSet")` objects is whether these objects are used to define a search space for tuning, or whether they are only used to check the validity of values. The `minsplit` parameter of the `r ref("mlr_learners_classif.rpart", "classif.rpart Learner")`, for example, has a lower but no upper bound. When tuning the value, however, a lower *and* upper bound must be given because tuning search spaces need to have finite measure. `r ref("ParamSet")` objects are used to describe both the search spaces of tuning runs, and the valid parameter ranges of `Learner`s and other objects. Therefore it is possible to create "unbounded" `r ref("ParamSet", text = "ParamSets")`, mostly in advanced usecases like creating or modifying `r ref("Learner")` or `r ref("PipeOp")` objects. Here, however, we will mainly focus on creating "bounded" `r ref("ParamSet", text = "ParamSets")` that can be used for tuning. See the [in-depth `paradox` chapter](#paradox) for more details on using `r ref("ParamSet", text = "ParamSets")` to  valid hyperparameter ranges for use-cases besides tuning.
```
jakob-r, line 5:
```suggestion
Note, that `r ref("ParamSet")` objects exist in two contexts.
First, `r ref("ParamSet")`-objects are used to define the space of valid parameter setting for a learner (and other objects).
Second, they are used to define a search space for tuning.
We are mainly interested in the latter.
For an example we can consider the `minsplit` parameter of the `r ref("mlr_learners_classif.rpart", "classif.rpart Learner").
The `r ref("ParamSet")` associated with the learner has a lower but *no* upper bound.
However, for tuning the value, a lower *and* upper bound must be given because tuning search spaces need to be bounded. 
For `r ref("Learner")` or `r ref("PipeOp")` objects, typically "unbounded" `r ref("ParamSet", text = "ParamSets")` are used.
Here, however, we will mainly focus on creating "bounded" `r ref("ParamSet", text = "ParamSets")` that can be used for tuning. 
See the [in-depth `paradox` chapter](#paradox) for more details on using `r ref("ParamSet", text = "ParamSets")` to  define parameter ranges for use-cases besides tuning.
```
one line per sentence would help here :)

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,223 @@
...
+    if (x$kernel == "polynomial") {
+      x$cost = x$cost * 2
+    }
+    x
+  }
+)
+rbindlist(generate_design_grid(tune_ps, 3)$transpose())
+```
+
+The available types of search space parameters is limited: continuous, integer, discrete, and logical scalars. There are many machine learning algorithms, however, that take parameters of other types, for example vectors or functions. These can not be defined in a search space `r ref("ParamSet")`, and they are often given as `r ref("ParamUty")` when defining a `r ref("Learner")`'s `r ref("ParamSet")`. When trying to tune over these hyperparameters, it is necessary to perform a Transformation that changes the type of a parameter.
```
jakob-r, line 88:
```suggestion
The available types of search space parameters are limited: continuous, integer, discrete, and logical scalars. There are many machine learning algorithms, however, that take parameters of other types, for example vectors or functions. These can not be defined in a search space `r ref("ParamSet")`, and they are often given as `r ref("ParamUty")` in the `r ref("Learner")`'s `r ref("ParamSet")`. When trying to tune over these hyperparameters, it is necessary to perform a Transformation that changes the type of a parameter.
```

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,223 @@
...
+  class.weights = p_dbl(0.1, 0.9, trafo = function(x) c(spam = x, nonspam = 1 - x))
+)
+generate_design_grid(tune_ps, 3)$transpose()
+```
+
+(We are omitting `rbindlist()` in this example because it breaks the vector valued return elements.)
+
+### Automatic Factor Level Transformation {#autolevel}
+
+A common use-case is the necessity to specify a list of values that should all be tried (or sampled from). It may be the case that a hyperparameter takes on function values and a certain list of functions should be tried. Or it may be that a choice of special numeric values should be used. For this, the `p_fct` constructor's `level` argument may be a value that is not a `character` vector, but something else. If, for example, only the values `0.1`, `3`, and `10` should be tried for the `cost` parameter, even when doing random search, then the following search space would achieve that:
```
jakob-r, line 103:
```suggestion
A common use-case is the necessity to specify a list of values that should all be tried (or sampled from). It may be the case that a hyperparameter accepts function objects as values and a certain list of functions should be tried. Or it may be that a choice of special numeric values should be tried. For this, the `p_fct` constructor's `level` argument may be a value that is not a `character` vector, but something else. If, for example, only the values `0.1`, `3`, and `10` should be tried for the `cost` parameter, even when doing random search, then the following search space would achieve that:
```

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,223 @@
...
+```{r 03-optimization-searchspace-011}
+tune_ps = ps(
+  cost = p_dbl(-1, 1, trafo = function(x) 10^x),
+  kernel = p_fct(c("polynomial", "radial")),
+  degree = p_int(1, 3, requires = kernel == "polynomial")
+)
+rbindlist(generate_design_grid(tune_ps, 3)$transpose(), fill = TRUE)
+```
+
+### Tuning ParamSets from other ParamSets
```
jakob-r, line 163:
```suggestion
### Create Tuning ParamSets from other ParamSets
```

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,223 @@
...
+
+Having to define a `r ref("ParamSet")` for a `r ref("Learner")` that already has parameter set information may seem unnecessarily tedious, and there is indeed a way to create tuning `r ref("ParamSet", "ParamSets")` from a `r ref("Learner")`'s `r ref("ParamSet")`, making use of as much information as already available.
```
jakob-r, line 165:
```suggestion
Having to define a tuning `r ref("ParamSet")` for a `r ref("Learner")` that already has parameter set information may seem unnecessarily tedious, and there is indeed a way to create tuning `r ref("ParamSet", "ParamSets")` from a `r ref("Learner")`'s `r ref("ParamSet")`, making use of as much information as already available.
```

bookdown/06-extending-learners.Rmd
```diff
@@ -230,6 +224,22 @@ For `classif.rpart` the following replace `<param_set>` above:
...
+        maxcompete = p_int(lower = 0L, default = 4L, tags = "train"),
+        maxsurrogate = p_int(lower = 0L, default = 5L, tags = "train"),
+        maxdepth = p_int(lower = 1L, upper = 30L, default = 30L, tags = "train"),
+        usesurrogate = p_int(lower = 0L, upper = 2L, default = 2L, tags = "train"),
+        surrogatestyle = p_int(lower = 0L, upper = 1L, default = 0L, tags = "train"),
+        xval = p_int(lower = 0L, default = 0L, tags = "train"),
+        keep_model = p_lgl(default = FALSE, tags = "train")
+      )
+```
+
```
jakob-r, line 242: Hmm, even if i like the shortcut way I don't know if we want to motivate people to use both ways when defining a learner. I am afraid of inconsistencies.

jakob-r (commented):
Nice write up!

bookdown/03-optimization-hyperband.Rmd
mb706: you mean the variable params? I can name it `tune_ps`, calling it `ps` would be confusing because of the `ps()` function we now have...

bookdown/06-extending-learners.Rmd
mb706: I believe it is useful for large parameter sets also when defining a Learner (or other object) ParamSet; most people will probably not write a learner to be committed to mlr3 for ever, but they will write something quickly to try something. And even for legitimate packages I think it is useful and people should be aware of the options: [This](https://github.com/mb706/mlrintermbo/blob/e0779569a110c6b4c559b77adfbcf9640fdc0e06/R/paramset.R) for example would have been about half as long using the shortcut notation.

mb706: What are concrete objections to `ps()` here? I can add them to the paragraph if there are pros and cons.

bookdown/03-optimization-hyperband.Rmd
jakob-r: `tune_ps` sounds perfect.

mb706: :+1:

bookdown/06-extending-learners.Rmd
jakob-r: Outside mlr3verse everyone is free to use whatever they want. My objections are less against `ps()` but more against doing the same thing in two different ways in terms of code style and maintainability. If there will ever be a change in the API we run into the risk in having to change two versions instead of just one. So we should just decide how to deal with it internally. As the book will be also read by "internal" people it might be worth to mention the code style - once we decided that.

bookdown/03-optimization-nested-resampling.Rmd
```diff
@@ -51,12 +51,18 @@ task = tsk("iris")
...
-at = AutoTuner$new(learner, resampling, measure = measure,
-  param_set, terminator, tuner = tuner)
+at = AutoTuner$new(
+  learner = learner, 
+  resampling = resampling, 
+  measure = measure,
+  terminator = terminator, 
+  tuner = tuner, 
+  search_space = search_space
+)
```
mb706, line 65: this is a bit verbose imho

jakob-r: i think @be-marc mainly changed it because the order of the arguments changed. You can shorten it I guess.

bookdown/06-extending-learners.Rmd
```diff
@@ -230,6 +224,22 @@ For `classif.rpart` the following replace `<param_set>` above:
       ps$values = list(xval = 0L)
 ```
 
+This can be written shorter, using the `r ref("paradox::ps")` shortcut:
```
jakob-r, line 227: How about:
```suggestion
Within mlr3 packages we suggest to stick to the lengthly definition for consistency, however the `<param_set>` can be written shorter, using the `r ref("paradox::ps")` shortcut:
```

jakob-r:
Is there anything blocking? If not, then @mb706 just decide how you want to deal with the remaining open points of the discussion and merge.

jakob-r (approved):

DESCRIPTION
```diff
@@ -69,6 +69,9 @@ Suggests:
     survAUC,
     svglite,
     xgboost
+Remotes:
+    mlr-org/mlr3tuning,
+    mlr-org/paradox
```
pat-s, line 74: Is this still required? IIRC we agreed on not using any dev versions in the book.

bookdown/03-optimization-searchspace.Rmd
```diff
@@ -0,0 +1,231 @@
...
+
+There are five domain constructors that produce a parameters when given to `r ref("ps")`:
+
+| Constructor               | Description                 | Is bounded?                | Underlying Class           |
+| :-----------------------: | :-------------------------: | :------------------------: | :------------------------: |
+| `p_dbl`                   | Real valued parameter ("double") | When `upper` and `lower` are given | `r ref("ParamDbl")` |
+| `p_int`                   | Integer parameter           | When `upper` and `lower` are given | `r ref("ParamInt")`    |
+| `p_fct`                   | Discrete valued parameter ("factor") | Always            | `r ref("ParamFct")` |
+| `p_lgl`                   | Logical / Boolean parameter | Always                     | `r ref("ParamLgl")`    |
+| `p_uty`                   | Untyped parameter           | Never                      | `r ref("ParamUty")`    |
```
pat-s, line 43: Can this table to formatted so that column widths match their contents? Many editors have auto-formatters for this.

bookdown/03-optimization-tuning.Rmd
```diff
@@ -9,7 +9,7 @@ This process is often called model 'tuning'.
 
 Hyperparameter tuning is supported via the `r mlr_pkg("mlr3tuning")` extension package.
 Below you can find an illustration of the process:
-
+2
```
pat-s, line 12: ?

pat-s (changes requested):
I mainly reviewed for style and consistency.

We are using the "one sentence per line" rule everywhere.
Would be great if this could be maintained in this PR as well.

DESCRIPTION
be-marc: We just do this inside a PR when the change is not on CRAN yet.

bookdown/06-extending-learners.Rmd
be-marc: I second that we stick to old code style for now. If somebody changes all learners to the new style, we can add this paragraph again.

PR #219, opened 2021-01-11
Fix file edit history URL

muskuloes:
The edit history URL didn't redirect properly. This PR fixes that.

Best :)


mllg:
Thanks!

PR #251, opened 2021-08-03
use the right CRS

jannes-m:
Hey @pat-s ,
the right CRS for the Ecuador dataset is in fact 32717 (see commit). Setting the CRS to 4326 just makes the dataset useless for geometry operations or plotting.
HTH,
Jannes


mllg:
@pat-s @be-marc can you check and merge?

pat-s (approved):

pat-s:
This was a mistake/typo by me, in {mlr3spatiotempcv} we use 32717 since the start. Thanks!

PR #252, opened 2021-08-10
remove redundant image link

mb706:
Change the end of 01: remove the part `To view the mlr3verse image...`; instead make the mlr3 ecosystem image itself into a link to itself and add the hint `click to enlarge`.


mllg:
I did the same in a different PR.

PR #374, opened 2022-06-21
Cleanup basics

mb706:


book/02-basics-tasks.qmd
```diff
@@ -108,22 +108,14 @@ The loading extension of packages can add to dictionaries such as `r ref("mlr_ta
...
-as.data.table(mlr_tasks)[, 1:4]
-```
-To get more information about a particular task, the corresponding man page can be found under `mlr_tasks_[id]`, e.g. `r ref("mlr_tasks_german_credit")`:
-```{r 02-basics-tasks-009, eval = FALSE}
-help("mlr_tasks_german_credit")
-```
-
-As an alternative, all `r mlr3book::mlr_pkg("mlr3")` objects come with a `help()` method which opens the corresponding help page.
-To open the help page of the previously-constructed palmer penguins task, you can also call:
+To get more information about a particular task, it is easiest to use the `help()` method that all `r mlr3book::mlr_pkg("mlr3")`-objects come with:
```
larskotthoff, line 111: Why is the example and mention of mlr3verse gone here?

mb706: 1. The text continues the sentence from the callout note, but I believe the callout notes should be more self-contained
2. The way that `library("mlr3verse")` is used here seems misleading to me, since someone who doesn't know what is happening  may think that some of the new tasks were introduced by mlr3verse, not by underlying packages.
3. I believe we should introduce the more user-friendly options first, so first `$help()`, then `help("<id of help page that is tedious to type and doesn't give tab completion or did-you-mean spelling correction>")`. The `help("...")` example is still here, just a few lines later 
4. We don't use the german credit task anywhere else on this page, this is why I switched it to penguins.

mb706: The german credit data is not even added by any extension package, it is in mlr3 just as well, so it doesn't demonstrate anything about needing to `library()` something else.

larskotthoff: Ok, I see your point. But mlr3verse should be introduced somewhere else then.

mb706: If you think we should "introduce" mlr3verse properly, then that should imho happen somewhere in 01-intro.qmd / 'introduction and overview'

larskotthoff (approved):

PR #394, opened 2022-08-05
remove unnecessary links

bblodfon:


mllg:
Thanks for your PR. Note that the pipelines chapter is undergoing a major rewrite (see #385), therefore I'm not sure if  changes to this chapter will remain in the book for long. Nevertheless, thanks! :)

PR #398, opened 2022-08-16
feat: add section about Filter-based feature selection

sebffischer:
* Fixes a typo
* Adds brief section about how to do filter-based feature selection

TODO: 
~~* [ ] Mention filter-based feature selection in the previous sections that list the different methods~~ (is already mentioned)

### DIFF
--- a/book/optimization.qmd
+++ b/book/optimization.qmd
@@ -1091,7 +1091,7 @@ head(as.data.table(filter), 3)
 ```
 ### Filter-based Feature Selection
 
-A common usecase for filters is to conduct feature selection based on the filter scores.
+A common use case for filters is to conduct feature selection based on filter scores.
 This can be achived using `r ref("mlr_pipeops_filter", text = "PipeOpFilter")`.
 This PipeOp takes as input a Task, applies the filter, and selects the features based on the calculated scores.
 The method how the features are subset can be defined in four different ways, each corresponding to a different parameter:
--- a/book/optimization.qmd
+++ b/book/optimization.qmd
@@ -1147,6 +1147,7 @@ rr = resample(task, lda_tuned, rsmp("cv"), store_models = TRUE)
 We can now extract the tuning results for each fold and visualize the dependency between the feature fraction and the classification error.
 
 ```{r optimization-062}
+library("ggplot2")
 archives = lapply(rr$learners,
   function(x) {
     archive = x$model$tuning_instance$archive
--- a/book/optimization.qmd
+++ b/book/optimization.qmd
@@ -31,6 +31,7 @@ In the [tuning](#tuning) part, we will have a look at:
 * [automating](#autotuner) tuning.
 
 We will use the `r mlr_pkg("mlr3tuning")` package, which supports common tuning operations.
+`
 
 **Feature Selection**
 
@@ -545,7 +546,7 @@ search_space = ps(
 rbindlist(generate_design_grid(search_space, 3)$transpose(), fill = TRUE)
 ```
 
-#G## Creating Tuning ParamSets from other ParamSets
+### Creating Tuning ParamSets from other ParamSets
 
 Having to define a tuning `r ref("ParamSet")` for a `r ref("Learner")` that already has parameter set information may seem unnecessarily tedious, and there is indeed a way to create tuning `r ref("ParamSet", "ParamSets")` from a `r ref("Learner")`'s `r ref("ParamSet")`, making use of as much information as already available.
 
@@ -1067,7 +1068,6 @@ filter_cor$param_set
 filter_cor$param_set$values = list(method = "spearman")
 filter_cor$param_set
 ```
-
 ### Variable Importance Filters {#fs-var-imp-filters}
 
 All `r ref("Learner")` with the property "importance" come with integrated feature selection methods.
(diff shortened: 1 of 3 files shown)


book/optimization.qmd
```diff
@@ -31,6 +31,7 @@ In the [tuning](#tuning) part, we will have a look at:
 * [automating](#autotuner) tuning.
 
 We will use the `r mlr_pkg("mlr3tuning")` package, which supports common tuning operations.
+`
```
mllg, line 34: ?

sebffischer:
ping

pat-s:
Danke, versuche es die Woche einzubauen.

book/optimization.qmd
```diff
@@ -1090,34 +1089,111 @@ filter = flt("importance", learner = lrn)
...
+  * `filter.cutoff` - All features with a filter score greater than the specified *cutoff* are selected.
+  * `filter.permuted` - A random permutation of each feature is added to the task before applying the filter.
+    All features selected before the *permuted*-th permuted features is selected are kept.
+
+
+:::{.callout-note}
+The `r ref("mlr_pipeops_filter", text = "PipeOpFilter")` only operates on the subset of features that have the correct column type and will leave other columns as is.
+:::
+
+We will illustrate the usage of this filter with the use of a machine learning pipeline for the `r ref("mlr_tasks_optdigits", text = "optdigits")` task, where the goal is to classify handwritten digits.
```
pat-s, line 1110:
```suggestion
We will illustrate the usage of this filter on a pipeline for the `r ref("mlr_tasks_optdigits", text = "optdigits")` task, where the goal is to classify handwritten digits.
```

book/optimization.qmd
```diff
@@ -1090,34 +1089,111 @@ filter = flt("importance", learner = lrn)
...
+We will first subset the features based on the `r ref("mlr3filters::mlr_filters_information_gain", text = "FilterInformationGain")` and then fit a `r ref("mlr3learners::mlr_learners_classif.lda", text = "LDA")`.
+We will tune the parameter `filter.frac` with a simple grid search and visualize the classification error for the different fractions.
```
pat-s, line 1112: Three times "we will" in a row, let's vary this a bit :)

pat-s, line 1112: At this point it would be good to mention that the tuning of filter hyperpars can be fusioned with learner hyperpar tuning. This is in fact one of the big advantages of filters: they don't need an extra layer of tuning (as wrappers do).

Maybe even worth putting this into a "tip" block.

book/optimization.qmd
```diff
@@ -1090,34 +1089,111 @@ filter = flt("importance", learner = lrn)
...
+
+```{r optimization-060}
+# reduce logging output
+lgr::get_logger("bbotk")$set_threshold("warn")
+lgr::get_logger("mlr3")$set_threshold("warn")
+
+task = tsk("optdigits")
+
+future::plan(future::multisession(workers = 4))
```
pat-s, line 1121: Do we want to have parallel code in the book?

book/optimization.qmd
```diff
@@ -1090,34 +1089,111 @@ filter = flt("importance", learner = lrn)
...
+
+lda_tuned = auto_tuner(
+  method = tnr("grid_search", resolution = 10),
+  learner = learner,
+  resampling = rsmp("holdout")
+)
+```
+
+:::{.callout-note}
+Unless you have a very specific reason, the parameter controlling how many features are selected, should be tuned.
```
pat-s, line 1138: Maybe phrase it a bit different: "outside of tutorials, [...]" 

The average user usually doesn't know whether they have a "very specific reason" :)

book/optimization.qmd
```diff
@@ -1090,34 +1089,111 @@ filter = flt("importance", learner = lrn)
...
+ggplot(data = archive, aes(x = as.factor(information_gain.filter.frac), y = classif.ce)) +
+  geom_boxplot() +
+  labs(
+    title = "Optdigits task: LDA for different subset fractions",
+    x = "Filter Fraction",
+    y = "Classification Error"
+  )
+```
+
+We can see that using 70% - 100% of the features seems to produce fairly similar results.
```
pat-s, line 1168: There's some more beefy content required here, this does not help the user much.
Maybe 2-3 sentences describing the results and then 1-2 interpreting/discussing them (e.g. "the learner seems to produce best results with most features present which means that only few noise is in the data and most features are important) etc.

It might be worth searching for a task in which feature selection actually has an effect on the performance (and the feature space), otherwise users might think: "what do I need it for, it does not make a difference". 
This might not be easy though and if we keep it like this, we need to explain to the user what the actual point is (and why we see 70-100% here and what that means). (Note: it also highly depends on the learner).

(I know it's an example and you know that, but in the book we should address this in a somewhat scientific detail, maybe even with some references :) )

pat-s (commented):
## Intro

A broader, more general introduction is needed which explains what filters and their "scores" are and how they differ compared to wrappers (one paragraph). 

## Conclusion

The following points should be briefly mentioned (maybe even in the intro, depends how you structure the section):

- Filters reduce the feature space and by that make models "simpler"
- They can be integrated into the tuning layer of the learner
- They can make use of caching and only need to be calculated once (large advantage compared to wrappers)

We should also mention ensemble filters, even though they are not yet available (only in the old mlr).

sebffischer:
I don't think we should mention the caching, because afaik we don't cache intermediate results in pipelines. 
The other things you suggest are already mentioned in the intro. 
In general I do not want to overdo it here, because it is a tutorial on usage. 
Also I don't think we should mention ensemble filters in the **mlr3** book if they are not avilable

pat-s:
> I don't think we should mention the caching, because afaik we don't cache intermediate results in pipelines.
The other things you suggest are partly already mentioned in the intro.
In general I do not want to overdo it here, because it is a tutorial on usage.
Also I don't think we need mention ensemble filters in the mlr3 book if they are not avilable.

This sounds like you don't really want to change anything substantial.

> I don't think we should mention the caching, because afaik we don't cache intermediate results in pipelines.

Apparently we don't have it yet in mlr3pipelines (https://github.com/mlr-org/mlr3pipelines/issues/16) which is somewhat of a pity as it is *very* important for filters.
This is one of the large benefits using filters over wrappers. 
AFAIR I didn't implement the same logic back as in {mlr} because we wanted to do it in pipelines but eventually this will never happen. 

> The other things you suggest are partly already mentioned in the intro.

This is quite general reply. I wouldn't have raised them if my feeling was they are appropriately covered.

> In general I do not want to overdo it here, because it is a tutorial on usage.

Sorry to be frank here but this sounds like an excuse for low motivation against medium complicated change requests to me ;)

> Also I don't think we need mention ensemble filters in the mlr3 book if they are not avilable.

The belong to the topic and are available in {mlr}, hence one sentence about them (whether they exist and/or why not) is certainly of interest to the reader.

---

I stand with my opinion that important information/content is missing or should be restructured and would not be happy seeing this getting merged in it's current form. If you do, I might want to rewrite it afterwards.

Please don't get this wrong, it's great we finally get a section about filters (and I could have done so earlier!) but if we have one, I'd like to see it being high quality and rather spent a bit more time on it.

sebffischer:
We discussed this a little today, and it was suggested to move the feature selection out of the optimization chapter.

RaphaelS1:
@sebffischer @be-marc how does this PR fit in with https://github.com/mlr-org/mlr3book/pull/412 ?

sebffischer:
> @sebffischer @be-marc how does this PR fit in with #412 ?

Thanks for the reminder. I moved the section into the chapter feature-selection. I guess the part is not perfect but now Marvin will rework this chaüter. Maybe we should just merge it and he can decide which parts to keep and what not?

RaphaelS1:
I think everything here is covered now in Chapters  5 and 6

PR #405, opened 2022-09-07
Remove quickstart section

jemus42:
To be replaced with some introductory statements by @berndbischl later.

### DIFF
--- a/book/index.qmd
+++ b/book/index.qmd
@@ -1,31 +1,7 @@
-# Quickstart {-}
+# The mlr3 book {-}
 
 {{< include _setup.qmd >}}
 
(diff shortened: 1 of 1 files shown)


jemus42:
Some introductory words / a "welcome" section for example, and/or acknowledgments - not entirely sure yet unfortunately, but Bernd said he'd write something.

Main thing was that the quickstart section itself seemed out of place and maybe not the best "first thing to see" for a new user opening the book. The content of the section itself seemed more appropriate for a package README rather than the foreword of a book.

larskotthoff:
Hmm, I would have preferred to have a PR that replaces it rather than removing it now and replacing it later at some point in time. But ok.

PR #459, opened 2022-11-08
Intro section for chapter 3

pfistfl:
@sumny could I get a quick review? 
Is this the depth we'd like to have for the intros?

This touches some of the same files as #419.
I'd suggest quickly merging this and then continuing with #419.

### DIFF
--- a/book/performance.qmd
+++ b/book/performance.qmd
@@ -3,18 +3,18 @@
 {{< include _setup.qmd >}}
 
 
-Now that we are familiar with the basics of how to construct tasks and learners and how to fit models them, we would like to know more about how to assess fitted models.
+Now that we are familiar with the basics of how to construct tasks and learners and how to fit models, we would like to know more about how to assess fitted models.
 In particular, we would like to assign one or several numeric *scores* to a given model.
 This allows us to select a *best* model out of several fitted models and furthermore, helps us assess whether the models we have fitted are useful for a task at hand.
 Scores in this context can reflect different qualities of a model, depending on the measures we employ.
 To provide an example, in some cases we might be interested in how well our model predicts (in terms of global accuracy).
 In others we might care about other perspectives, such as how many features it uses or how fast it predicts.
 
-In supervised learning contexts, the core idea in performance evaluation is often that we would like to assess how good a fitted model will work on new data points.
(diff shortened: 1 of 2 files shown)


sumny (commented):
In general I would say this is the depth we should aim for!
One could think about introducing some formalism here but I am not sure if this helps here that much.

book/performance.qmd
```diff
@@ -3,34 +3,63 @@
 {{< include _setup.qmd >}}
 
 
-Now that we are familiar with the basics of how to create tasks and learners, how to fit models, let's have a look at some of the details, and in particular how `r mlr_pkg("mlr3")` makes it easy to perform many common machine learning steps.
+Now that we are familiar with the basics of how to construct tasks and learners and how to fit models them, we would like to know more about how to assess fitted models.
```
sumny, line 6: typo: fit models them

book/performance.qmd
```diff
@@ -3,34 +3,63 @@
...
+In particular, we would like to assign one or several numeric *scores* to a given model.
+This allows us to select a *best* model out of several fitted models and furthermore, helps us assess whether the models we have fitted are useful for a task at hand.
+Scores in this context can reflect different qualities of a model, depending on the measures we employ.
+To provide an example, in some cases we might be interested in how well our model predicts (in terms of global accuracy).
+In others we might care about other perspectives, such as how many features it uses or how fast it predicts.
 
-We will cover the following topics:
+In supervised learning contexts, the core idea in performance evaluation is often that we would like to assess how good a fitted model will work on new data points.
+In order to mimic this, we often hold out some data during model development. 
+Afterwards, we use predictions made on the held-out data and compare them to the true label to simulate how our model would have fared on new data points.
```
sumny, line 15: maybe even use "new, unseen data"

book/performance.qmd
```diff
@@ -3,34 +3,63 @@
...
+This strategy is often called *train-test* split or *holdout*.
+[Performance Measures](#measures)now summarize such comparisons in a single number, e.g. a model's accuracy is the fraction of cases in which it predicted the true outcome correctly.
```
sumny, line 17: space missing: Performance Measures](#measures) now

PR #480, opened 2022-11-25
preamble proofreading

larskotthoff:


book/index.qmd
```diff
@@ -1,30 +1,31 @@
 # Preamble {.unnumbered}
 {{< include _setup.qmd >}}
 
-```{r C0 seed}
+```{r C0 seed, echo = FALSE}
```
RaphaelS1, line 4: I think we do want to echo this. People need to be able to reproduce everything

book/index.qmd
```diff
@@ -1,30 +1,31 @@
...
 ```{r C0 install mlr3, eval = FALSE}
 install.packages("mlr3")
 ```
 
-In this first example we'll show you the most basic use-case, train and predict.
+In our first example, we will show you some of the most basic functionality -- training a model and making predictions.
 
 ```{r C0 egBasic}
-library("mlr3")
+library(mlr3)
```
RaphaelS1, line 25: Happy with this change but need to make sure we do this everywhere -- I've been changing as I see them

book/index.qmd
```diff
@@ -35,9 +36,19 @@ predictions
...
-Whilst mlr3 makes training and predicting easy, it also uses a unified interface to perform some very complex operations in just a few lines of code:
+* load the `r mlr3` library,
+* choose the 'penguins' "task" (the `r mlr3` term for a dataset),
+* randomly split the task into 67% training data and 33% testing data (this is the default when using the `partition()` function),
+* choose the `rpart` decision tree learner to perform classification,
+* train the model on the task, selecting the training rows from the split,
+* show the fitted model,
+* make predictions for test rows in the penguins task,
+* show the predictions, and
+* compute the accuracy performance measure for those predictions, i.e. the fraction of predictions that are correct.
```
RaphaelS1, line 49: Personally I'm not sure the list works here as the point I want to make is that we can do easy things and complex things - but when we write the above as a long list it makes it seem more difficult.

book/index.qmd
```diff
@@ -35,9 +36,19 @@ predictions
 predictions$score(msr("classif.acc"))
 ```
 
-Here we have picked the 'penguins' task (which is `r mlr3` language for dataset), randomly split the task into 67% training data and 33% testing data, trained a random forest on the training data to learn the probability of an observation falling into one of the outcome classes, showed the fitted model, and then made prediction on the test data, showed these predictions and evaluated the model using the accuracy measure.
+We
```
RaphaelS1, line 39:
```suggestion
In this example, we
```

book/index.qmd
```diff
@@ -72,38 +83,57 @@ bma
...
+In this (much more complex!) example we chose two tasks and two machine learning algorithms ("learners" in `r mlr3` terms).
+We used automated tuning to optimise the number of trees in the random forest learner (@sec-optimization) and a machine learning pipeline that imputes missing data, collapses factor levels, and creates stacked models (@sec-pipelines).
+We also showed basic features like loading learners (@sec-basics) and choosing resampling strategies for benchmarking (@sec-performance).
+Finally, we compared the performance of the models using the mean accuracy on the test set, and applied a statistical test to see if the learners performed statistically significantly different (they did not!).
```
RaphaelS1, line 89: I don't think we need double definition of 'statistical' so suggest revert back to '...performed significantly different..."

book/index.qmd
```diff
@@ -72,38 +83,57 @@ bma
...
+We recommend that all readers study these chapters to become familiar with basic mlr3 terminology, syntax, and style.
+@sec-optimization, @sec-feature-selection, and @sec-pipelines contain more advanced implementation details and some machine learning theory.
+@sec-special delves into detail on domain-specific methods that are implemented in our extension packages.
+Readers may choose to selectively read sections in this chapter depending on your use cases (i.e., if you have domain-specific problems to tackle), or to use these as introductions to new domains to explore.
+@sec-technical contains technical implementation details that are essential reading for advanced users who require parallelisation, custom error handling, and fine control over hyperparameters and large databases.
+Many readers will likely find that the built-in model interpretability methods will fit the majority of needs, however readers that require more detailed methods for interpreting a model (for example those training deep neural networks or gradient boosting machines) will find @sec-interpretation very useful.
+Finally, anyone who would like to contribute to our ecosystem should read @sec-extending.
```
RaphaelS1, line 110: Thanks for changing to dynamic references, not sure why I hard-coded!!

book/index.qmd
```diff
@@ -72,38 +83,57 @@ bma
...
+This book tries to follow the Diátaxis framework for documentation and so we include tutorials, how-to guides, API references, and explanations.
+This means that the conclusion of each chapter includes a short reference to the core functions learnt in the chapter, links to relevant posts in the [mlr3gallery](https://mlr-org.com/gallery.html), and a few exercises that will cover content introduced in the chapter.
+You can find the solutions to these exercises in @sec-solutions.
 
-Finally, if you want to reproduce any of the results in this book, firstly note the seed set at the top of each chapter (which are simply set as the chapter number) and the `sessionInfo` printed in @sec-session-info.
+Finally, if you want to reproduce any of the results in this book, note that the random seed is set as the chapter number and the `sessionInfo` printed in @sec-session-info.
```
RaphaelS1, line 122: BTW I know here we state the seed is chapter number and that's an argument to set `echo=FALSE` above but my guess is not everyone will read this which is why I think we should echo

book/index.qmd
```diff
@@ -112,35 +142,43 @@ options(repos = c(
...
 
 ## Community links
 
-The mlr community is open to all and we welcome all people, from those completely new to machine learning and R up to advanced coders and data scientists. You can talk to us at anytime by joining our [Mattermost](https://lmmisld-lmu-stats-slds.srv.mwn.de/signup_email?id=6n7n67tdh7d4bnfxydqomjqspo).
+The mlr community is open to all and we welcome everybody, from those completely new to machine learning and R to advanced coders and professional data scientists.
+You can reach us on our [Mattermost](https://lmmisld-lmu-stats-slds.srv.mwn.de/signup_email?id=6n7n67tdh7d4bnfxydqomjqspo).
 
-For case-studies and how-to guides, check out [https://mlr-org.com/gallery.html](https://mlr-org.com/gallery.html) for extended practical blog posts. Or for mlr updates you might find our blog a useful point of reference [https://mlr-org.com/blog.html](https://mlr-org.com/blog.html).
+For case studies and how-to guides, check out [the mlr gallery](https://mlr-org.com/gallery.html) for extended practical blog posts.
```
RaphaelS1, line 165: This doesn't work for people reading a published book which is why I've changed all links to explicitly show the URL.

book/index.qmd
```diff
@@ -112,35 +142,43 @@ options(repos = c(
...
+For updates on mlr you might find [our blog](https://mlr-org.com/blog.html) a useful point of reference.
```
RaphaelS1, line 166: As above

book/index.qmd
```diff
@@ -112,35 +142,43 @@ options(repos = c(
...
 
-with `{pkg}` replaced by the package you want to install. You can see an up-to-date list of all our extension packages at [https://github.com/mlr-org/mlr3/wiki/Extension-Packages](https://github.com/mlr-org/mlr3/wiki/Extension-Packages).
+with `{pkg}` replaced with the name of the package you want to install.
+You can see an up-to-date list of all our extension packages at [https://github.com/mlr-org/mlr3/wiki/Extension-Packages](https://github.com/mlr-org/mlr3/wiki/Extension-Packages).
 
 ## Community links
 
-The mlr community is open to all and we welcome all people, from those completely new to machine learning and R up to advanced coders and data scientists. You can talk to us at anytime by joining our [Mattermost](https://lmmisld-lmu-stats-slds.srv.mwn.de/signup_email?id=6n7n67tdh7d4bnfxydqomjqspo).
+The mlr community is open to all and we welcome everybody, from those completely new to machine learning and R to advanced coders and professional data scientists.
+You can reach us on our [Mattermost](https://lmmisld-lmu-stats-slds.srv.mwn.de/signup_email?id=6n7n67tdh7d4bnfxydqomjqspo).
```
RaphaelS1, line 163: Arguably this one is too complicated to write-out the full URL and so I'd suggest we use bitly or cutly

book/index.qmd
```diff
@@ -112,35 +142,43 @@ options(repos = c(
...
+For case studies and how-to guides, check out [the mlr gallery](https://mlr-org.com/gallery.html) for extended practical blog posts.
+For updates on mlr you might find [our blog](https://mlr-org.com/blog.html) a useful point of reference.
 
-We appreciate of all contributions to our codebase, whether they are bug reports, feature requests, or pull requests. Each of GitHub repositories includes issues and pull request templates to ensure we can help you as much as possible to work together. Please make sure you read our code of conduct and contribution guidelines to keep our community friendly and safe! With many packages in our universe it may be hard to keep track of where to open issues. As a general rule:
+We appreciate all contributions, whether they are bug reports, feature requests, or pull requests that fix bugs or extend functionality.
+Each of our GitHub repositories includes issues and pull request templates to ensure we can help you as much as possible to get started.
+Please make sure you read our code of conduct and contribution guidelines to keep our community friendly and safe!
```
RaphaelS1, line 170: Need to add these

RaphaelS1:
Okay how's this for that last line:

"When we write out functions for the first time we will include all parameter names and default arguments, for example `benchmark(design,  store_models = FALSE,  store_backends = TRUE,  encapsulate = NA_character_,  allow_hotstart = FALSE,  clone = c("task", "learner", "resampling"))`. However when using functions in code chunks we only include parameter names for optional paramters, for example

```Rbenchmark(

We do not include required function parameter names in code chunks, only optional ones. But we always include the full function signature the first time it is introduced. For example say we are introducing a new function called adding_function:

File: book/index.qmd (hunk 1 above)
Comment (RaphaelS1) [reply book/index.qmd:4]:
```suggestion
```{r C0 seed}
```

.github/workflows/build-book.yml
```diff
@@ -47,6 +47,9 @@ jobs:
           cd book
           R -q -e 'options(Ncpus = 2); install.packages("renv", repos = "cloud.r-project.org"); renv::restore()'
 
+      - name: Document and install book
+        run: make bookinstall
```
RaphaelS1, line 51: @be-marc please note this change I've made to the workflow, it broke before because the book wasn't being documented and installed so new R functions were'nt registered. I don't *think* this breaks the changes you made but also I don't know how `renv` works so please fix if needed!

RaphaelS1 (approved):

PR #462, opened 2022-11-10
refactor: basics intro

sumny:
Not sure if we want to introduce some concepts in more depth here, e.g., hypothesis space, empirical risk minimization, model parameters vs. learner hyperparameters; could also introduce some formalism but not sure if this actually helps the reader..

@pfistfl could you do a brief review?

PR #497, opened 2022-12-19
add latex-math

sumny:
* `input` statements go in `book/_quarto.yml`
* note that some macros may need changes due to conflicts with `\bm`, `\boldsymbol` and unicode-math

### DIFF
--- a/book/_quarto.yml
+++ b/book/_quarto.yml
@@ -32,7 +32,7 @@ book:
     - feature-selection.qmd
     - pipelines.qmd
     - preprocessing.qmd
-    - special.qmd
+    - special.qnd
     - technical.qmd
     - interpretation.qmd
     - extending.qmd
@@ -87,6 +87,7 @@ format:
     code-block-bg: true
     code-overflow: wrap
     code-line-numbers: true
+    pdf-engine: pdflatex
     include-in-header:
       text: |
         \usepackage{makeidx}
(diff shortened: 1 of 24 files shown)


sumny:
* `\renewcommand` in macros apparently only works properly if we use `pdflatex`
* but switching to `pdflatex` apparently breaks rendering of paradox symbols: `LaTeX Error: Unicode character ∈ (U+2208)`
* trying to fix this with `inputenc` package will result in clash due to `makeidx`
* Maybe @RaphaelS1 can review it and decide how to move forward

RaphaelS1:
I don't really know what I'm looking at, these are just tex files, what's the application to the book?

sumny:
this is needed to have consistent notation in the intros of (at least) chapter 2, 3, 4, because @berndbischl wanted to have some formal introduction as well.

relevant for at least @giuseppec and me but maybe also others.

I just wanted to provide a separate PR so that the individual persons who rely on the tex macros in their intros can merge it own their own in their branch (or we merge it directly in main)

RaphaelS1:
Do you have an example of where this is used?

sumny:
> Do you have an example of where this is used?

see e.g., #462 the intro (still very drafty and too much formalism but according to @berndbischl we should use some notation here)

RaphaelS1:
I think we need to discuss off GH with @berndbischl. To me it doesn't make sense to add this much latex when we aren't using most of it, and from what I can tell there's not enough in the intro to warrant the pain with the macros, we could just use the full commands. I'm also not sure if Basics makes the most sense for the maths, if we put it into the Intro chapter then it's more clearly related to all subsequent chapters (which imo should have 0 maths but only code)

sumny:
as we decided to NOT add that much math, no longer needed

Issue #264, opened 2021-10-04
doc package options on cheatsheet / book


berndbischl:
from package man page

"mlr3.debug": If set to TRUE, parallelization via future is disabled to simplify debugging and provide more concise tracebacks. Note that results computed with debug mode enabled use a different seeding mechanism and are not reproducible.

"mlr3.allow_utf8_names": If set to TRUE, checks on the feature names are relaxed, allowing non-ascii characters in column names. This is an experimental and temporal option t

also check for other stuff

RaphaelS1:
Do we still have/use these options @berndbischl @mllg ?

mllg:
These two options are still in use. I can add a note to the parallelization chapter.

RaphaelS1:
Thanks! Would you mind adding a similar short note to basics maybe (or whether you yhink best) for the second?

mllg:
Yep, I will add this to PR #501.

PR #501, opened 2023-01-08
Add additional note about mlr3.debug flag

mllg:
For #264.

Additional note about utf8 names still missing.


book/basics.qmd
```diff
@@ -107,6 +107,12 @@ task_mtcars = as_task_regr(mtcars_subset, target = "mpg", id = "cars")
 task_mtcars
 ```
 
+:::{.callout-tip}
+The task constructors `r ref("as_task_regr()")` and `r ref("as_task_classif()")` will check for non-ascii characters in the column names of your data.
+As many ML models do not work properly with arbitrary [UTF8 names](https://en.wikipedia.org/wiki/UTF-8), `r mlr3` defaults to throw an error if any of the column names contains either a non-ascii character or does not comply to R's variable naming scheme.
+We generally recommend converting names with `r ref("make.names()")` first, but if you are feeling lucky you can also set the option `mlr3.allow_utf8_names` to true to relax the check.
```
RaphaelS1, line 113:
```suggestion
We generally recommend converting names with `r ref("make.names()")` first, but you can also set the option `mlr3.allow_utf8_names` to true to relax the check (but do not be surprised if a model fails due to variable naming).
```

RaphaelS1 (approved):

Issue #503, opened 2023-01-11
Chapter 4v3 comments


be-marc:
Version 2 of Chapter 4 Optimization  is merged to main. Please add your comments for version 3 here.

RaphaelS1:
Amazing thanks @be-marc ! Some final comments and then I think we should consider the chapter 'locked' for further editing until others are ready.

* [ ] The solution to exercise 3 is missing
* [ ] Can you add a small amount of text (1-2 sentences) at start of 4.2 to explain why it's 'Advanced' 
* [ ] There are broken references, please go through and check these, e.g. '?sec-fallback-learner'
* [ ] Caption for Fig 4.5 needs updating with new style
* [ ] For Fig 4.6 please can you add some text into the diagram itself that clearly says something like "The wrong way to tune"
* [ ] Can you see if you can find a way to fix the content (i.e., column sizes) in table 4.3?

be-marc:
>  Can you see if you can find a way to fix the content (i.e., column sizes) in table 4.3?

The table with s3 and r6 classes? We cannot break `mlr3tuning::extract_inner_tuning_archives()`. But yes the summary column looks silly.

be-marc:
![grafik](https://user-images.githubusercontent.com/33069354/212668708-a8fd7428-926e-4cd5-b1ee-fb11867d064f.png)

Since we stopped using the logscale this plot is pretty useless isn't it?

RaphaelS1:
> The table with s3 and r6 classes? We cannot break mlr3tuning::extract_inner_tuning_archives(). But yes the summary column looks silly.

Yeah don't worry for now, I'm going to remove pkg prefixes on Fri

> Since we stopped using the logscale this plot is pretty useless isn't it?

Agreed but it is nice to show and explain that plot. Anywhere else we could have a version of the plot?

be-marc:
> Agreed but it is nice to show and explain that plot. Anywhere else we could have a version of the plot?

Maybe we should add logscale to the search space again?

RaphaelS1:
> Maybe we should add logscale to the search space again? Logscale is now in the last section. Before the plot is useless.

Could we move the plot later on and reference it earlier?

larskotthoff:
At the very beginning of the chapter, the seed is set without any explanation. Briefly explain why the seed should be set and how reproducibility is affected (especially important for tuning).

[...]

PR #507, opened 2023-01-15
Optimization v3

be-marc:
**Discussion in #503**


book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+
+An example is the `class.weights` parameter of the [Support Vector Machine](https://machinelearningmastery.com/cost-sensitive-svm-for-imbalanced-classification/) (SVM), which takes a named vector of class weights with one entry for each target class.
+The transformation that would tune `class.weights` for the `r ref("mlr_tasks_spam", text = "spam")` dataset could be:
+
+```{r optimization-041}
+search_space = ps(
+  class.weights = p_dbl(lower = 0.1, upper = 0.9,
+    trafo = function(x) c(spam = x, nonspam = 1 - x))
+)
+```
```
RaphaelS1, line 885: I think this is a bit confusing following from the previous example (also subsequent examples below). Why sometimes `.extra_trafo` and others `trafo`. Why not use vector example again? (these are rhetorical Qs I know the answer) I think the flow could be a bit clearer, maybe something like "We will now look at common advanced tuning methods" then you can talk about vectors, characters, etc. But just framed as problem/solution

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+This is equivalent to the following:
+```{r optimization-043}
+search_space = ps(
+  cost   = p_fct(c("0.1", "3", "10"),
+    trafo = function(x) list(`0.1` = 0.1, `3` = 3, `10` = 10)[[x]]),
+  kernel = p_fct(c("polynomial", "radial"))
+)
+```
+
+This may seem silly, but makes sense when considering that factorial tuning parameters are always `character` values:
```
RaphaelS1, line 909: Never say "this may seem silly" about your own code

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+typeof(search_space$params$cost$levels)
+```
+
+:::{.callout-warning}
+Be aware that this results in an "unordered" hyperparameter.
+Tuning algorithms that make use of ordering information of parameters, like genetic algorithms or model based optimization, will perform worse when this is done.
+For these algorithms, it may make more sense to define a `r ref("p_dbl")` or `r ref("p_int")` with a more fitting transformation.
+:::
+
+It is possible to specify dependencies between parameters.
```
RaphaelS1, line 925: This has already been discussed above

book/optimization.qmd
```diff
@@ -113,25 +113,27 @@ We do this by constructing a learner and using `r ref("to_tune()")` to set the l
...
+  cost  = to_tune(1e-1, 1e5),
+  gamma = to_tune(1e-1, 1),
   type  = "C-classification",
   kernel = "radial"
 )
 learner
 ```
 
 Here we have constructed a classification SVM by setting the type to "C-classification", the kernel to "radial", and not fully specifying the `cost` and `gamma` hyperparameters but instead indicating that we will tune these parameters.
+The section ... and ... give
```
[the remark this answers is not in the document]
be-marc: Forgot the ref

book/optimization.qmd
```diff
@@ -113,25 +113,27 @@ We do this by constructing a learner and using `r ref("to_tune()")` to set the l
...
 ::: {.callout-note}
 The `cost` and `gamma` hyperparameters are usually tuned on the logarithmic scale.
-You can find out more in @sec-advanced-search-spaces.
+You can find out more in @sec-logarithmic-transformations.
 :::
 
 Search spaces are usually chosen by experience.
-In some cases these can be quite complex, @sec-paradox describes how to construct these.
+In some cases these can be quite complex.
+The sections X and Y give a more detailed insight into the creation of tuning spaces.
```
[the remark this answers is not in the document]
be-marc: Added ref

book/optimization.qmd
```diff
@@ -282,11 +285,16 @@ instance = tune(
 instance$result
 ```
 
+::: {.callout-note}
+The measured performance is different from the previous section because not the same train-test splits were used.
+The train-test splits are generated at the beginning of the optimization with the current seed (see @sec-resampling-inst).
+:::
+
```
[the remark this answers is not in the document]
be-marc: @larskotthoff @berndbischl Your suggestion

book/optimization.qmd
be-marc: @sebffischer

book/optimization.qmd
be-marc: @sebffischer

book/optimization.qmd
be-marc: @sebffischer

book/optimization.qmd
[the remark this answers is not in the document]
larskotthoff: I think there's value in at least briefly mentioning this.

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+learner$param_set$search_space()
+```
+
+A `r ref("TuneToken")` can also be constructed with a `r ref("Domain")` object, i.e. something constructed with a `p_<type>` call.
+This way it is possible to tune continuous parameters with discrete values, or to give transformations or dependencies.
+One could, for example, tune the `cost` as above on three given special values, and introduce a dependency of `shrinking` on it.
+Notice that a short form for `to_tune(<levels>)` is a short form of `to_tune(p_fct(<levels>))`.
+
+:::{.callout-note}
+When introducing the dependency, we need to use the `degree` value from *before* the implicit transformation, which is the name or `as.character()` of the respective value, here `"val2"`!
```
sebffischer, line 763:
```suggestion
When introducing the dependency, we need to use the `cost` value from *before* the implicit transformation. Concretely, we write the dependency as `cost == "val2"` and not `cost == 0.7`.
```

book/optimization.qmd
sebffischer, line 906:
```suggestion
Keep in mind, that factorial tuning parameters are always `character` values:
```

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+:::
```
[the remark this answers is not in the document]
sebffischer: Done

book/optimization.qmd
sebffischer: Maybe a graphic would be nice, which illustrates how the trafos are first applied and then the .extra_trafo is applied to all the transformed parameters?

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+```{r optimization-029}
 rr$aggregate()
 ```
 
 As a final note, nested resampling is computationally expensive, as a simple example using five outer folds and three inner folds with a grid search of resolution 5 used to tune 2 parameters, results in 5*3*5*5 = 375 iterations of model training/testing. In practice, you may often see closer to three folds used in inner resampling or even holdout, or if you have the resources then we recommend parallelization (@sec-parallelization).
 
+## Advanced Tuning
+
+This section is devoted to advanced tuning techniques.
+The topics are not necessary for a basic understanding of tuning but make tuning more efficient and robust.
```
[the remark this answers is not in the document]
RaphaelS1: @be-marc please add the optional section tag or don't and resolve this conversation instead

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+
+### Encapsulation and Fallback Learner {#sec-encapsulation-fallback}
+
+So far, we have only looked at the case where no issues occur.
+However, it often happens that learners with certain configurations do not converge, run out of memory, or terminate with an error.
```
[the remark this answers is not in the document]
RaphaelS1: I'm merging this suggestion as it's literally repeating the line before.

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+)
+
+learner$encapsulate = c(train = "evaluate", predict = "evaluate")
+```
+
+The encapsulation can be set individually for training and predicting.
+There are currently two options for encapsulating a learner.
+The  `r ref_pkg("evaluate")` package and the `r ref_pkg("callr")` package.
+The `r ref_pkg("evaluate")` catches any errors that occur and allows the process to continue.
+The `r ref_pkg("callr")` package comes with more overhead because the encapsulation spawns a separate R process.
```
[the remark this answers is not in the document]
RaphaelS1: @be-marc I don't think this has been addressed

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+```{r optimization-031}
+learner$timeout = c(train = 30, predict = 30)
+```
+
+With encapsulation, exceptions and timeouts do not stop the tuning.
+Instead, the error message is recorded and a fallback learner is fitted.
+
+Fallback learners allow scoring a result when no model was fitted during training.
+A common approach is to predict a weak baseline e.g. predicting the mean of the data or just the majority class.
+See @sec-fallback for more detailed information.
```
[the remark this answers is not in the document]
RaphaelS1: @be-marc this hasn't been addressed - where is sec-fallback?

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+  measures = msr("classif.ce"),
+  term_evals = 10
+)
+
+as.data.table(instance$archive)[, list(cost, gamma, classif.ce, errors, warnings)]
+```
+
+### Logarithmic Transformations {#sec-logarithmic-transformations}
+
+Usually, the `cost` and `gamma` hyperparameters are tuned on the logarithmic scale which means the optimization algorithm searches in $[log(1e-5), log(1e5)]$ but transforms the selected configuration with `exp()` before passing to the learner.
```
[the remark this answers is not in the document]
RaphaelS1: @be-marc has this been addressed?

book/optimization.qmd
```diff
@@ -772,12 +592,404 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+instance$result
+```
+
+The column `x_domain` contains the hyperparameter values after the transformation i.e. `exp(5.76)` and `exp(-5.76)`:
+
+```{r optimization-054}
+instance$result$x_domain
+```
+
+### Creating Search Spaces from Learners {#sec-search-space-token}
```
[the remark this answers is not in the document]
RaphaelS1: @be-marc

book/optimization.qmd
RaphaelS1: @be-marc

book/optimization.qmd
RaphaelS1: @be-marc

RaphaelS1 (changes requested):
@be-marc quite a few comments are unresolved, can you take a quick look?

book/optimization.qmd
sebffischer: I added an explanation

book/optimization.qmd
sebffischer: I removed this repetition

book/optimization.qmd
```diff
@@ -772,12 +591,402 @@ It is therefore important to ensure that the performance of a tuned model is *al
...
+
+The search space picks up dependencies from the underlying `r ref("ParamSet")` automatically.
+So if the `kernel` is tuned, then `degree` automatically gets the dependency on it, without us having to specify that.
+
+```{r optimization-049}
+learner$param_set$values$kernel = to_tune(c("polynomial", "radial"))
+learner$param_set$search_space()
+```
+
+It is even possible to define whole `r ref("ParamSet")`s that get tuned over for a single parameter.
```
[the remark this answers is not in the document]
sebffischer: Ok this is not necessary because it is decently explained I think

book/optimization.qmd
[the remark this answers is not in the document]
sebffischer: @be-marc this has nothing to do with paradox

book/optimization.qmd
RaphaelS1: It's still confusing but is made clearer with the example

book/optimization.qmd
```diff
@@ -660,7 +663,11 @@ as.data.table(instance$archive)[, list(cost, gamma, classif.ce, errors, warnings
 
 ### Logarithmic Transformations {#sec-logarithmic-transformations}
 
-Usually, the `cost` and `gamma` hyperparameters are tuned on the logarithmic scale which means the optimization algorithm searches in $[log(1e-5), log(1e5)]$ but transforms the selected configuration with `exp()` before passing to the learner.
+After the previous section has covered how to make the tuning process more robust to failures, we will now show how to exercise more fine-grained control over the sampling distribution.
```
RaphaelS1, line 666:
```suggestion
We will now look at another advanced topic, which is using transformations to exercise more fine-grained control over the sampling distribution.
```

book/optimization.qmd
```diff
@@ -660,7 +663,11 @@ as.data.table(instance$archive)[, list(cost, gamma, classif.ce, errors, warnings
...
+In `r ref_pkg("mlr3")`, the standard way to modify the sampling distribution -- besides changing the tuner from e.g. grid search to random search -- is to apply a transformation.
```
RaphaelS1, line 667:
```suggestion
```

book/optimization.qmd
```diff
@@ -660,7 +663,11 @@ as.data.table(instance$archive)[, list(cost, gamma, classif.ce, errors, warnings
...
+To show how this works, we use the SVM's `cost` and `gamma` parameters that we have tuned in the example above.
+These are usually not tuned on a linear, but logarithmic scale.
```
RaphaelS1, line 669:
```suggestion
To show how this works, we use the SVM's `cost` and `gamma` parameters that we have tuned in the example above, these are usually tuned on a logarithmic (as opposed to linear) scale.
```

book/optimization.qmd
```diff
@@ -728,9 +735,10 @@ instance$result$x_domain
...
 ### Creating Search Spaces from Learners {#sec-search-space-token}
 
-We will now explain in more detail how a search space is created from a Learner.
-This is done by setting values of a `r ref("Learner")`'s `r ref("ParamSet")` to so-called `r ref("TuneToken")`s, constructed with a `r ref("to_tune")` call.
-This can be done in the same way that other hyperparameters are set to specific values.
+In @sec-tuning-instance we have seen how one can conveniently define the tuning space of a learner using a `r ref("TuneToken")` that is constructable with the `r ref("to_tune")` function.
+In this section, we will provide a more detailed explanation of the underling mechanism and its capabilities.
+
+Hyperparameters can be set to the `r ref("TuneToken")` in the same way that they are set to specific values.
 It can be understood as the hyperparameters being tagged for later tuning.
```
RaphaelS1: I don't know what this means sorry

book/optimization.qmd
```diff
@@ -838,6 +846,7 @@ This transformation can access all parameter values of an evaluation and modify
 It is even possible to add or remove parameters.
 
 In our example we now assume that the parameter `cost` should be set to a higher value when the `kernel` is `"polynomial."`.
+Such a transformation is only possible via a `.extra_trafo`, because it accesses information from multiple parameters at once.
```
RaphaelS1, line 849: Have we defined `.extra_trafo` already? If not this won't make sense

book/optimization.qmd
```diff
@@ -783,7 +783,10 @@ learner$param_set$search_space()
 
 It is even possible to define whole `r ref("ParamSet")`s that get tuned over for a single parameter.
 This may be especially useful for vector hyperparameters that should be searched along multiple dimensions.
-This `r ref("ParamSet")` must, however, have a `.extra_trafo` that returns a list with a single element, because it corresponds to a single hyperparameter that is being tuned.
+This can be done using an `.extra_trafo`, which is similar to the `logscale` transformation from earlier, but is applied to all parameters.
+It can be set to a function with arguments `x` and `param_set` -- the parameter values and the `r ref("ParamSet")` itself -- and must return the transformed parameters.
+You will learn more about it in section @sec-paradox.
+In this case special case, the `.extra_trafo` must return a list with a single element, because it corresponds to a single hyperparameter that is being tuned.
```
RaphaelS1, line 789:
```suggestion
In this special case, the `.extra_trafo` must return a list with a single element, as it corresponds to a single hyperparameter that is being tuned.
```

book/optimization.qmd
```diff
@@ -798,6 +801,9 @@ learner$param_set$set_values(class.weights = to_tune(par))
 learner$param_set$search_space()
 ```
 
+:::{.callout-warning}
```
RaphaelS1, line 804:
```suggestion
::{.callout-tip}
```

PR #514, opened 2023-01-20
Fix ref function - Closes #500

### LINKED ISSUE #500: The `ref` function needs fixing
At least it should have an argument "package" to link to external packages and should not print the `mlr3<package>::` prefix unless explicitly stated to agree with the mlr3book style guide outlined in the preface

RaphaelS1:
@sebffischer please can you review and merge?

Closes #500


RaphaelS1:
@be-marc would appreciate if you could check I updated lockfile correctly

be-marc:
Looks good.

be-marc:
The other packages were also updated because they are remotes in `mlr3book`.

Issue #439, opened 2022-10-21
Chapter 2 comments


RaphaelS1:
## Style

* [x] This could be massively shortened by combining all code in one chunk than explaining it in text
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L533-L550
* [x] Consider `lm` over `rpart` as model output is considerably less and therefore neater
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L527
* [x] Don't list all learners and measures just the `head`
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L454-L461
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L689

## Content

* [x] Because we distinguish between Tasks and Data in mlr, consider splitting them in the diagram. I don't think Repeat/Resample is needed at this stage, it's a bit confusing to look at without context (it isn't clear how this relates to using the same or different data/measure/model etc.). Also move figure after the text at the end of the section preamble.
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L10
* [x] Need a clear discussion between difference between learner and model, currently I think a bit too brief.
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L16
* [x] Remove, not useful
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L25
* [x] This needs more consistency. If you're going to give a quick overview then make sure it's detailed enough to be understood, e.g. "unsupervised task to estimate the density" hasn't explained the task. Consider removing tasks in development.
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L66
* [x] Too technical. This is covered in a later chapter
https://github.com/mlr-org/mlr3book/blob/ac3f3b95eee0f4d1556734d83d05860596ad4c16/book/basics.qmd#L104
* [x] Either in Chapter 1 or 2 (maybe 1) need to explain dictionaries
[...]

PR #516, opened 2023-01-23
Added section on thresholding for multiple classes

nfoss2:


nfoss2:
This PR addresses comment from issue: #439

larskotthoff:
Thanks, could you add some more detail please? In particular, what does the threshold number mean and why might you want to adjust it?

PR #525, opened 2023-02-09
Technical v2

mllg:


mllg:
> Parallelization
> 
>     * I think the entry to the chapter is in general very useful, but a bit too technical as the intro I think.. Maybe we could instead start with a more accessible, concrete parallelization example and use this to explain concepts like “main”, “threads” and “forks”?
> 

I've tried both orders, I preferred the current.

>     * “We call the parallelization platform together with the respective interface provided by an R package the parallelization backend.” What is the parallelization platform?

Rephrased.


> 
>     * The random forest example is used twice (maybe this is also ok)

Unsure where :(

> 
>     * Why are no timings shown for the mclapply example?

Switched to a socket example, hopefully this will not break the CI or give random numbers in the book.

> 
>     * Parallel prediction is missing

Yep ... 

> 
>     * Can we somewhere add a graphic that illustrates the parallelization overhead (including the writing and reading of `.rds` files etc.) ?

I'm not creative enough for this one :)

> 
> Error Handling:
> 
>     * Maybe the “Fallback Learner” section would be better suited in the “Large-Scale Benchmarking Experiments” (?) At least the part where we talk about how to address the analysis of such results as this is methodological and not technical. Maybe we can also simply duplicate this and reference the sections appropriately.
> 

Needs discussion with Bernd ord Raphael.

> 
> Logging:
> 
>     * Do we maybe want this already earlier in the book (is this even mentioned somewhere (?)) Because disabling the logger might be something that people want to be able to do before they arrive at the technical chapter.
> 

IMHO the logger is not too verbose for most users, except for tuning. Maybe just throw in some references in the respective chapters?

mllg:
> Data Backends:
> 
>     * I am not sure whether we really want to show how to work with the DBI package here. People should go to the respective package vignettes for that (?)
>       →I think we should focus more on the mlr3::DataBackend instead.
>       We first create a sqlite file and then create a sqlite connection to manipulate data. Someone who does not know this, should instead do the calculation using data.table and then use the provided `as_sqlite_backend()` converter. And someone who already knows how to do this with the DBI package does not need the explanation either

I get you point, but would also like to discuss with Bernd and Raphael.

> 
>     * The first paragraph of the Parquet / DuckDB Chapter applies to the whole chapter. Why is this in the subsection?

Good point, moved the paragraph up.

mllg:
> Parameters:
> 
>     * Most of this is moved to other chapters
> 
>     * I moved the chapter as a paradox vignette, to keep the information that is not moved to other chapters. The PR is here: https://github.com/mlr-org/paradox/actions/runs/4529676246/jobs/7977681421?pr=390

Most or all of it?

mllg:
> * Maybe it makes sense to already introduce the debug learner in the parallelization section, because it is useful to play around with parallelization by setting the sleep_train parameter? Maybe we can just switch the order of sections 1 and 2 and this problem would be solved.

I'd rather focus on parallelization there, but combining this might be a nice exercise :)



> 
> * The references table is missing
> 
> * Exercises are missing

Yep. I'll write down one now.

> 
> * The conclusion is missing

This chapter is a mixed bag, not sure what to write here. Is this mandatory?

> 
> * The “Further resources” section is missing (I am not sure whether we even want this consistently?)

sebffischer:
Sorry I did not mean the random forest example, but the example with parallelizing the split points, but I think it is no problem that it is mentioned twice. 

> This chapter is a mixed bag, not sure what to write here. Is this mandatory?

I think we want it for consistency. The conclusion then also contains the reference table e.g. 

> I'm not creative enough for this one :) (a graphic for the parallelization overhead)

Maybe something like the [image](https://www.mcs.anl.gov/~itf/dbpp/text/img384.gif) from [here](https://www.mcs.anl.gov/~itf/dbpp/text/node29.html) + some description of the figure?

> IMHO the logger is not too verbose for most users, except for tuning. Maybe just throw in some references in the respective chapters?

Yes, I think references are enough.

> Most or all of it?

Most of it. Some stuff still has to be added to the extending chapter, and other things should not be covered in the book I think. The parts that will not be contained in the book anymore are how the samplers (grid sampler, uniform sampler) work, or how to create ParamSets without sugar. I think it is a good solution to point to the paradox vignette where all of this is covered in more detail.

sebffischer:
Summary comment with TODOs / things to talk about in the next meeting: 
Edit this from now on when adding new TODOs etc. 
If we not want this, then cross it out instead of removing it / ticking it off. 

* [  ] Do references / a conclusion make sense for this chapter. 
* [  ] Add exercises for other chapter. Currently there is only one for the parallelization. 
* [ ] Merge the paradox vignette to preserve information that will not be in the book anymore. PR is open [here](https://github.com/mlr-org/paradox/pull/390)
* [ ] Add section about further resources
* [ ] Add section about parallel prediction
* [ ] Add references in the earlier chapters to inform the users how to modify the logger (e.g. for tuning)
* [ ] Discuss with Bernd / Raphael whether it makes sense to move some of the error handling / fallback learner to "Large Scale Experiments" chapter
* [ ] Discuss with Bernd / Raphael whether we want to keep the data manipulation with DBI or focus more on the `mlr3::DataBackend` itself. 
* [ ] Check once again whether we removed anything from paradox that we want to keep in technical (?)

larskotthoff:
Have all comments been addressed here? It looks like there are still some open issues in the latest preview (e.g. mlr_measures$mget --> msrs(), no resources section, only the one exercise).

mllg:
Still unsure what to write in intro, abstract and conclusion, but maybe we can still push this forward to the next review stage. I think I've resolved most or all of the other stuff.

PR #537, opened 2023-02-28
Modifying `performance-043` chunk.

rchemist0123:
To run the `performance-043` chunk properly, the `predict_type="prob"` argument is needed. Moreover, `rr1` is modified as `rr`.


jemus42:
Thanks for catching that! 
That chapter (and many others) is currently undergoing a rewrite, see #520.
I'll make sure to fix that up there, but will close this PR to centralise on the refactoring branch.

PR #552, opened 2023-03-16
Optimization v4

be-marc:


sumny:
currently I don't like: intro is after all these reiterations very very basic and not even the concept of GE is mentioned anymore still we use the term sometimes; maybe see how this is introduced in chapter 3 and resolve this

larskotthoff:
- [ ] Probably not necessary to have the code to set the seed show up in the book. If we keep it, there should be some explanation.
- [ ] Would mention at the beginning that parameters determine the model and hyperparameters determine the way the model is built to make this connection between them more explicit.
- [ ] Would also mention at the beginning that methods like BO are adaptive, i.e. later iterations depend on what happened in earlier ones.
- [ ] Figure 4.1 looks weird in the PDF, e.g. the i in Configuration is missing. Something gone wrong with the conversion?
- [ ] Colors in Figure 4.1. aren't clear. I think blue and purple are swapped? Some objects belong to different categories, which makes it more unclear. I would just highlight the user-supplied objects in the figure with a different color.
- [ ] 4.1.1 in the example at the end for constructing a learner with `to_tune()` it might be worth pointing out that this learner will give an error message when trained as a "normal" learner.
- [ ] 4.1.2 explain the meaning of the `threshold` argument for stagnation.
- [ ] I would move 4.1.4 Tuner immediately after 4.1.1 Learner and Search Space. Feels weird about terminating without knowing yet what will be terminated, especially with references to terminator none for grid search.
- [ ] Code example in 4.1.3 `resampling` and `measure` aren't used.
- [ ] Table 4.2 shows all packages as mlr3*, which seems to suggest that all tuners are implemented by the mlr3 team. Would make more sense to list the underlying package there, and mlr3tuning only if the particular tuner is actually implemented in there.
- [ ] Table 4.2 caption mentions "methodology", but the table doesn't seem to contain that.
- [ ] 4.1.4 paragraph about grid search doesn't explain how values for categorical parameters are chosen if the resolution is less than the total number of values.
- [ ] 4.1.4 paragraph on grid search I would replace "naive" with "non-adaptive".
- [ ] 4.1.4 in the paragraph on BO I would mention that the acquisition function also (potentially) uses the uncertainty of predictions of the surrogate model.
- [ ] 4.1.4 in the paragraph on multi-fidelity HPO I would make the connection to race where evaluation on subsets of folds is already mentioned (and maybe that's a better place to talk about racing).
- [ ] 4.1.4 paragraph on multi-fidelity HPO mentions anytime performance without explaining it.
- [ ] 4.1.4 is there a way to determine automatically what tuner works on what space? Section talks about hierarchical spaces, non-numeric HPs etc which not all tuners can handle. Can a user find the list of "applicable" tuners automatically, similar to learners that can handle a task? If so, this should be mentioned, if not, could think about implementing this (but maybe not for this book).
- [ ] 4.1.4 paragraph on choosing the right tuner should briefly mention grid search with a coarse resolution and why that might not be a good idea.
- [ ] 4.1.4 same paragraph might be worth mentioning that if the budget is extremely small (e.g. only the number of initial samples for BO), random search usually does at least as well as adaptive methods. In particular suitable for quickly trying HPO.
- [ ] End of 4.1.4 mentions `batch_size` without details -- should at least mention that depending on the tuner values larger than 1 might not "work".
- [ ] ParamSet at the end of 4.1.4 should mention what `param_resolutions` means.
- [ ] 4.1.5 difference between `x_domain` and `learner_param_vals` is unclear -- sounds like `learner_param_vals` is a strict superset? Then why store both? Might be helpful to show both.
- [ ] 4.1.6 mention briefly why a log scale makes more sense for these hyperparameters than a linear scale. At least point out for the example that we got a much better result than when tuning on a linear scale before.
- [ ] 4.1.7 text says with `tune()` I have less control, but code seems to do exactly the same thing. Give example where I can't use `tune()`.
- [ ] 4.1.8 would say a little bit more about how the plot is generated, in particular that this requires training a surrogate model (which learner is used for that?).
- [ ] 4.1.8. might be worth adding an example that uses the same space, but BO and then plot the result to show that it focuses on the area where the lowest error is.
- [ ] 4.2 can I get the archive from an autotuner and analyze the tuning result in the same way as shown before?
- [ ] 4.3 would be good to have a little bit more explanation at the beginning, e.g. the usual thought experiment about HPs that don't do anything but there's random noise in the evaluation.
- [ ] Figure 4.4 colors in the figure and caption are different.
- [ ] 4.3 at the end of the description of nested resampling it might be worth mentioning explicitly what the "output" of each level is -- inner resampling produces generalization performance estimates based on which a single HP configuration is chosen to be evaluated on the outer resampling; outer resampling produces generalization estimates for these optimal HP configs. In particular the outer resampling does not produce an optimized HP config. Might also be helpful to say how many times total HPO will be run in this example.
- [ ] 4.3.1 at the end say explicitly how HP config to fit a final model should be obtained.
- [ ] 4.3.2 mentions overfitting without explaining it.
- [ ] 4.3.2 is unclear to me -- inner performances 0.23 0.15 0.19 don't look significantly better than outer performances 0.17 0.17 0.23. Should also point out that the inner resampling results are aggregated over the inner folds while the outer ones are individual performance numbers. Text then says that significant difference indicates overfitting, but when aggregating the outer fold performance estimates we somehow get an unbiased estimate?
- [ ] 4.3.3 "Instructive Example" -> "Detailed Example"?
- [ ] 4.3.3 introduces data generators and `tgen()` without explanation. Briefly describe.
- [ ] 4.3.3 none of the code examples produce output, but it looks like some of them should.
- [ ] Unsure about 4.3.3 in general. Nested resampling only makes a very small difference. Maybe kNN and tuning k would show a larger difference? At the moment I'm not sure if this example adds anything useful -- the previous section already shows that there are (small) differences in the performance estimates. An additional example would need to show a much larger difference.
- [ ] 4.4.1 in the first code example say explicitly that "evaluate" arguments refer to the evaluate package. What are other possible arguments, e.g. if I want to use callr?
- [ ] 4.4.1 mention why one might want to use callr -- the text makes it sound like there's no advantage as it adds additional overhead.
- [ ] 4.4.1 what happens if I don't set a fallback learner?
- [ ] 4.4.1 would be good to show an example where an error actually happens. Maybe subset sonar to contain only one class? That might cause some errors? Or set a really low timeout.
- [ ] 4.5.1 what happens when a search space is not bounded? Error message? Infinite loop?
- [ ] 4.5.1 for `ParamUty` it might be worth pointing out that while this cannot be in the *search space* paramset, it can be in the paramset of a learner to be tuned.
- [ ] 4.5.1 don't really understand the example with the "numeric" factors (and the list of values is different in different code snippets) -- is the equivalent code with the trafo function generated automatically or how does the coercion to/from characters happen? So is the `p_fct(c(0.3, 0.7))` essentially syntactic sugar?
- [ ] 4.5.2 should this be labelled advanced as well?
- [ ] 4.5.2 don't understand the initial example. Why can I not just say `depends = cost == 0.3`? Why do I need the designation as a special named value? The output says "Trafo is set." but the code doesn't set a trafo.
- [ ] 4.5.3 says tuning spaces are taken from publications that are encoded in the name, but doesn't list the publications -- where do "default" and "rbv2" spaces come from? Is the list of publications somewhere in the documentation?
- [ ] 4.6.1 not sure if this section is necessary. We don't give that much detail on any of the other methods, and it's not clear that this level of details is beneficial for the reader. I would remove this section and simply point to the paper.
- [ ] 4.6.2 would be shorter to use `lts()` to retrieve the default space and modify only the `nrounds` HP.
- [ ] 4.6.2 r_min and r_max are defined both when the HP space is created and when HB is instantiated. Are both necessary? Also should `hyperband_schedule()` be `tuner$hyperband_schedule()`? Or is this call not actually modifying anything but only illustrating what the schedule looks like? If so say that explicitly.
- [ ] 4.6.2 "This does not necessarily have to be a configuration evaluated with the highest budget since XGBoost can in general overfit the data when using too many boosting iterations." doesn't make sense to me. Does this mean that XGBoost automatically detects overfitting and some how prevents HB from evaluating with the highest budget?
- [ ] 4.6.3 would swap that with 4.6.2 so that the "familiar" example of tuning the SVM comes first. Or actually maybe this should be moved to the pipelines chapter.
- [ ] 4.6.3 could we use `lts()` to get this search space and only modify the budget param?
- [ ] 4.6.3 same as in 4.6.2 -- do the budget ranges need to be specified twice?
- [ ] 4.6.3 text says that best model is SVM with radial kernel, but code output shows poly kernel.
- [ ] Figure 4.9 looks like there's only a single blue dot? How can the number of selected features be fractional? I would move the reference to this figure to the end of the section and in the beginning only say that there will be a figure to illustrate Pareto front.
- [ ] 4.7 difference between Pareto set and Pareto front not entirely clear -- I would phrase this as the Pareto set defining the Pareto front, along which we can get the values for points not in the Pareto set or something like this.
- [ ] 4.7 the best configurations shown at the end of the section have each exactly the same values for both objectives. Might be better to have an example where the result is more "interesting".
- [ ] Table 4.5 R6 class names protrude into the summary column in the PDF.

be-marc:
Moved to #579

PR #558, opened 2023-03-23
Feature Selection v3 (Chapter 5)

mnwright:
Small changes after 2nd review with @berndbischl.


larskotthoff:
- The paragraph at the end of the introduction ("the reader should know") refers to an appendix (which the reader almost certainly hasn't read yet). We don't have this for other chapters -- I would simply omit this paragraph. If somebody hasn't read the previous chapters, it's on them if they don't understand something.
- In the introduction, I would make the fundamental difference between types of filtering methods clear -- those that do not use any information from a learner that will be trained on the selected features (learner-agnostic) and those that do (learner-specific). This is already there implicitly, but I think it would help to make it explicit.
- 5.1.1 I would add a code example that shows the filter values with spearman.
- 5.1.2 why does the code example remove missing data?
- 5.1.3 in the code example it would be good to select the learner that supports `selected_features` in the same way as in 5.1.2, i.e. list the learners that support this instead of `stopifnot()`, which doesn't really help the user.
- Description at the start of 5.1.4 is misleading -- `$select()` allows to choose the feature names to be kept, not specifically the ones whose score is 1.
- 5.2 I would briefly mention at the beginning that this is a way to get learner-specific information from learners that don't explicitly support it through the methods mentioned before.
- In the code example in 5.2.1, it might be worth pointing out that `fselect` instantiates and runs the feature selection, as opposed to the tuning just before, where the instance needs to be created and the optimization process run explicitly.
- When showing the result of this code example (the archive), it might be worth explaining the meaning of the `batch_nr` column and at least pointing out that there are more batches. This confused me because the plot immediately afterwards is showing much more information than the table.
- 5.2.2 for forward selection, can a terminator be supplied? E.g. automatically terminate when all combinations have been tried, or a timeout is reached? The description also sounds like technically `TerminatorStagnation` is the default when nothing is specified; is this the case?
- 5.2.3 might be worth pointing out that the main difference of the different feature selection algorithms is the order in which feature subsets are tried. Parts of the current description (e.g. for exhaustive search) sound like the terminator is ignored.
- 5.2.4 code example for showing the best feature subset is truncated (...); would be good to show how to get a list of selected features.
- 5.2.4 code example that shows the archive only shows values for some features, leading to identical rows with different accuracy in the output. Maybe run the example on a subset of features for penguins to avoid this? Would also solve the issue above.
- 5.2.5 suggest adding a plot of the Pareto front.
- 5.2.6 might be worth pointing out the differences to a pipeline.

mnwright:
Thanks Lars! I'll work on your comments/suggestions next week.

mnwright:
Thanks again! I'm done but have a few comments on your suggestions: 

> The paragraph at the end of the introduction ("the reader should know") refers to an appendix (which the reader almost certainly hasn't read yet). We don't have this for other chapters -- I would simply omit this paragraph. If somebody hasn't read the previous chapters, it's on them if they don't understand something.

The reference to the appendix was a mistake (we had a duplicate #sec-tasks); fixed now. I like the paragraph of what readers should know, but if we don't have that anywhere else, I'm also fine with removing it. 

> In the introduction, I would make the fundamental difference between types of filtering methods clear -- those that do not use any information from a learner that will be trained on the selected features (learner-agnostic) and those that do (learner-specific). This is already there implicitly, but I think it would help to make it explicit.

Good point. Actually no filter methods are learner-specific, as they can always combined with any learner afterwards. I've made a few changes to the text and added a note box explaining this. 

> 5.2 I would briefly mention at the beginning that this is a way to get learner-specific information from learners that don't explicitly support it through the methods mentioned before.

Since we don't really have learner-specific methods (see above), I think that would lead to confusion. But I've added a note similar to that in filters to explain that here the learner is *not* independent. 

> 5.2.5 suggest adding a plot of the Pareto front.

I like the idea but I don't see how that could be done. We have a {0,1}^60 space (each of the 60 features is in or out) which I have no idea how to plot.

larskotthoff (approved):

PR #560, opened 2023-03-27
[Review only] Comments spatial

pat-s:


book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% rsmp_sp = rsmp("repeated_spcv_coords", folds = 3, repeats = 2, id = "SpCV")
+%
+% design = benchmark_grid(task, learner, c(rsmp_nsp, rsmp_sp))
+% bmr = benchmark(design)
+% bmr$aggregate(msr("classif.acc"))[, c(5, 7)]
+% ```
+%
+% In the above example, the non-spatial resampling method means that the train and test data are very similar due to spatial autocorrelation, and so this is almost as bad as evaluating a model on the training data (see @sec-basics).
+% In contrast, the spatial method has accommodated for autocorrelation and the test data is therefore less correlated (though still some association will remain) with the training data.
+% Visually this can be seen using built-in `r ref("autoplot")` methods.
```
pat-s, line 951:
```suggestion
% This can be seen visually by using the built-in `r ref("autoplot")` methods of `r mlr3spatiotempcv`.
```

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% In @fig-sprsmp we visualize how the task is partitioned according to the spatial resampling method (@fig-sprsmp top) and non-spatial resampling method (@fig-sprsmp bottom).
+% There is a clear separation between coordinates using the 'correct' spatial resampling method whereas the train and test splits overlap significantly (and are therefore highly correlated) using the non-spatial method.
```
pat-s, line 953:
```suggestion
% There is a clear separation in space for the respective partitions when using the spatial resampling whereas the train and test splits overlap a lot (and are therefore more correlated) using the non-spatial method.
```

Given that "significantly" is often associated with a statical meaning I'd suggest to avoid the term in such contexts.

The degree of correlation depends and is not always "high".

"seperation between coordinates" is uncommon wording here. One would rather say "Separation in space for the respective partitions".

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% #| label: fig-sprsmp
+%
+% library(patchwork)
+% autoplot(rsmp_sp, task, fold_id = c(1:3), size = 0.7) /
+%   autoplot(rsmp_nsp, task, fold_id = c(1:3), size = 0.7) &
+%   scale_y_continuous(breaks = seq(-3.97, -4, -0.01)) &
+%   scale_x_continuous(breaks = seq(-79.06, -79.08, -0.02))
+% ```
+%
+% Now we have seen why spatial resampling methods matter we can take a look at what is implemented in `r mlr3spatiotempcv`.
```
pat-s, line 967:
```suggestion
% After we have seen why spatial resampling matters, we can take a look at what is implemented in `r mlr3spatiotempcv`.
```

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% At the time of writing we have implemented 19 different resampling methods for spatial and spatiotemporal analysis.
```
pat-s, line 968: I'd rather list the unique ones and not count Repeated/non-repeated as a unique method.

Also I'd like to point out that "we" didn't implement/invent them but rather collected them from various packages/implementations out there. Big difference :)

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% These are categorized into:
+%
+% * Blocking -- Create rectangular blocks in 2D/3D space (as opposed to partitioning in 1D space used in non-spatial methods)
```
pat-s, line 971: All spatial methods operate at least in 2D space. This sounds like as if only "blocking" would do.

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% * Buffering -- Create buffering zones to remove observations between train and test sets
+% * Spatiotemporal clustering -- Clusters based on coordinates (and/or time-points)
+% * Feature space clustering -- Clusters based on feature space and not necessarily spatiotemporal
+% * Partitioning -- Grouped by factor variables
```
pat-s, line 975: This description would optimally align with the categorization we did in https://arxiv.org/abs/2110.12674 where we used "Custom (partitioning)" for the factor var grouping and also added more information about each category.

Also partly available at https://mlr3spatiotempcv.mlr-org.com/articles/mlr3spatiotempcv.html#upstream-packages-and-scientific-references.

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+%
+% The choice of method may depend on specific characteristics of the dataset and there is no easy rule to pick one method over another, full details of different methods can be found in @Schratz2021 -- the paper deliberately avoids recommending one method over another as the 'optimal' choice is highly dependent on the predictive task, autocorrelation in the data, and the spatial structure of the sampling design.
+% The documentation for each of the implemented methods includes details of each method as well as references to original publications.
+%
+% :::{.callout-tip}
+% ## Spatio*temporal* resampling
+%
+% We have focused on spatial analysis but referred to "spatiotemporal" and "spatiotemp".
+% The resampling methods discussed in this section extend trivially to temporal analysis (or spatial and temporal analysis combined) by setting the `"time"` `col_role` in the task (@sec-performance).
```
pat-s, line 984:
```suggestion
% The spatial-only resampling methods discussed in this section extend trivially to temporal analysis (or spatial and temporal analysis combined) by setting the `"time"` `col_role` in the task (@sec-performance).
```

pat-s: Reading this as a user, I would be a bit confused:

- What does it mean here "extend trivially"? What would I need to do exactly? (Disclaimer: I don't think it's "that easy")
- Defining `time` as a col_role is only helpful (and needed) when aiming for a spatiotemporal partitioning, which is quite advanced as the partitioning gets a lot more complex.

I'd suggest to reword and take out "trivally" and mention that so far everything shown has focused on spatial-only partitioning and extending it to spatiotemporal is possible but advanced.

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% See the `r mlr3spatiotempcv` visualization `r link("https://mlr3spatiotempcv.mlr-org.com/articles/spatiotemp-viz.html", "vignette")` for specific details about 3D spatiotemporal visualization.
+% :::
+%
+% ### Spatial Prediction {#sec-spatial-prediction}
+%
+% Until now we have looked at resampling to accommodate spatiotemporal *features*, but what if you want to make spatiotemporal *predictions*?
+% In this case the goal is to make classification or regression predictions for spatial coordinates at a pixel level, usually by creating and labelling raster images.
```
pat-s, line 991:
```suggestion
% In this case the goal is to make classification or regression predictions at the pixel level, i.e. for an area, defined by the geometric resolution, of a raster image.
```

"Spatial coordinates at pixel level" doesn't go together in this context as a pixel is composed out of multiple coordinates (4) and you would not predict just these and then interpolate. I.e. you don't do "point predictions" but "area predictions" here and one pixel = 1 prediction, 1 coordinate != 1 prediction.

book/special.qmd
```diff
@@ -851,188 +851,188 @@ The experiment shows that using the K-means algorithm with four centers has the
...
+% lrn = lrn("classif.ranger")$train(tsk_leipzig)
+% pred = predict_spatial(tsk_predict, lrn, format = "terra")
+% pred
+% ```
+%
+% In this example we specified creation of a `terra` object, which can be visualized with in-built plotting methods.
+%
+% ```{r special-050, message = FALSE, cache.lazy=FALSE, cache=FALSE}
+% library(terra, exclude = "resample")
+% plot(pred, col = c("#440154FF", "#443A83FF", "#31688EFF",
```
pat-s, line 1033:
```suggestion
% autoplot(pred, col = c("#440154FF", "#443A83FF", "#31688EFF",
```

pat-s (commented):
Thanks for the update/rewrite of the section last month. It's always good to get external input and views on existing content!
I think though that we would have profited from discussing some high-levels ideas/changes upfront before doing a semi-hard rewrite of a chapter which somebody else composed initially.

I'd also highly favor having the authors being attached to the individual subsections, *especially* in the "special" chapter as there are so many different applications in it. And I think it would be much better for readers (and authors) to know how wrote what. The current state makes this very unclear as all are grouped at the top of the chapter.

A lot of wording used in the section is somewhat uncommon WRT to the spatial field/community and tenses seem to be a bit off in some/many occasions (I haven't labeled all cases).

## General comments

- Usually we (the field) prefer to say "spatial autocorrelation" instead of "autocorrelation in spatial data" due to the many manifestations of autocorrelation, i.e. the former makes it more clear that it is about this specific type while the latter one is more tending towards "it could be any type of autocorrelation in spatial data". 
- Many sentences are very long and could/should be split in two or three sentences.
- I suggest to use more generic wording like "spatial resampling" instead of "spatial resampling methods" which is unnecessary verbose and sounds strange in many sentences (just one example, there are more occasions of this claim).
- I think the parallelized spatial prediction feature of {mlr3spatial} (which also Marc has put a lot of time into) could be highlighted more/at all in the prediction subsection.
- I see the spatiotemp visualization got removed completely. While I guess it's because of 3D graphics, there are ways to include 2D static snapshots in a print version of the book. Seeing it not being mentioned at all given that the content already existed feels a bit sad.

## Technical

If you want to apply some suggestions directly, you can just do so in the PR and before merging remove all the `%` prefixes again from the content to have a clean merge.

larskotthoff:
Thanks Patrick! It looks like something went wrong with the PR though -- all lines have a % at the beginning, which breaks the formatting. Did you mean to comment/remove those lines?

pat-s:
This is intended and explained in the review comment.

RaphaelS1:
https://github.com/mlr-org/mlr3book/pull/562

Issue #561, opened 2023-03-29
Chapter 8 comments


larskotthoff:
- [x] The point of the paragraph "So far, this book has focused on detailed explanation of the [mlr3](https://mlr3.mlr-org.com/) universe of packages, which is abstracted in such a way that allows users to choose the level of complexity that suits each project." is unclear to me. The text suggests that if you're using only mlr3 you can only do small (sub)sets of data. The whole paragraph seems redundant.
- [x] Second paragraph mentions "deep learning architectures, image analysis, multi-label classification" as supported, but doesn't give any details on how to do this. I would move this to the end of the chapter and provide pointers to relevant packages. Packages "waiting to be developed" shouldn't be mentioned; otherwise we might as well say that, taking future development into account, mlr3 will be the best and most popular ML package ever.
- [x] Not sure about the fictional company example in 8.1. An insurance company wouldn't give out loans, and we don't have similarly "chatty" examples elsewhere.
- [x] In the benchmarking example in 8.1.1 I would not use `paste0` and simply spell out all the learners. This makes it unnecessarily difficult to understand, especially as `paste0` isn't explained.
- [x] Worth pointing out at the end of 8.1.1 that a positive loss means a loss for the company, i.e. none of the models performs as well as we would like to.
- [x] 8.2. By example -> For example
- [x] 8.2.1 "Predicting a confidence interval is possible using only features in [mlr3](https://mlr3.mlr-org.com/)" -- don't understand this. Why would it not be possible using only features? What else could we use to make the prediction?
- [ ] 8.2.1 Where does the 95% confidence interval come from? Is this arbitrary, is it what the `se` prediction type is always guaranteed to return? Also briefly mention relationship to `quantile = 1.96`.
- [ ] "Analogously to classification, we might think of the top prediction as “P(Y = 30) = 0.4” versus the fourth prediction as “P(Y = 23) = 0.7” (numbers not exact). This level of detail allows more nuanced metrics to be utilized to measure probabilistic predictions" -- not clear to me what this means. I can do the same thing without confident intervals. Why does it help to think of predictions this way?
- [ ] Example in 8.2.2 should continue the one above (i.e. mean and se from a real model) instead of arbitrary numbers.
[...]

PR #562, opened 2023-03-30
Special v3

RaphaelS1:
@larskotthoff I've addressed all comments from https://github.com/mlr-org/mlr3book/issues/561 in https://github.com/mlr-org/mlr3book/pull/562/commits/4254916064493d31ed9692f4b3e55a5fec84a4dc (Note ignored comments about Prob Regr for now subject to conversation in MM)

@pat-s I've addressed all comments from https://github.com/mlr-org/mlr3book/pull/560 in https://github.com/mlr-org/mlr3book/pull/562/commits/3f7d405ef43c28033986b1f48d5cda07aeefc90a

Re-reviews appreciated


larskotthoff (approved):

larskotthoff:
Thanks, looks good. I'll have a look at the entire chapter again once this has been merged.

book/special.qmd
```diff
@@ -878,27 +751,31 @@ x1 = mean(m$lon)
...
+  labs(fill = "Cases")
 
-p1 / p2
+p1 / p2 & patchwork::plot_layout(guides = "collect")
 ```
 
-Unlike other tasks we have looked at in this chapter, there is no underlying difference to the implemented learners or measures, instead we provide new resampling methods in `r mlr3spatiotempcv` to account for any overconfidence in predictions that might occur from autocorrelation.
-To make use of these resamplings we have also implemented special tasks in `r mlr3spatial` to hold spatial data.
+Unlike other tasks we have looked at in this chapter, there is no underlying difference to the implemented learners or measures.
+Instead we provide new resampling methods in `r mlr3spatiotempcv` to account for the similarity in the train and test sets during resampling that originates from spatial autocorrelation.
```
pat-s, line 778:
```suggestion
Instead we provide additional resampling methods in `r mlr3spatiotempcv` to account for the similarity in the train and test sets during resampling that originates from spatiotemporal autocorrelation.
```

The methods are not new, i.e. all of them existed before and we just "use" them. There is no method that has been newly developed in tempcv.

I think adding a disclaimer here regarding the use of the words "spatial autocor." or "spatiotemporal autocor." would be good. I would prefer to use/introduce the acronym STAC (spatiotemporal autocorrelation) and state that it depends on the data and context whether the autocorrelation is spatial, temporal or spatiotemporal.
This way one does not always have to switch between the different terms in each section and it is clear that the focus is on accounting for a combination of all, if existent.

EDIT: I see there is a disclaimer/note coming further down the text. I'd move it up so readers see it earlier and get clarification on the use of the words and different types of autocor.

book/special.qmd
```diff
@@ -928,12 +805,12 @@ Once a task is created, you can train and predict as normal.
...
+However as discussed above, it is best to use the specialized resampling methods to achieve bias-reduced estimates of model performance.
 
 ### Spatiotemporal Cross-Validation {#spatiotemp-cv}
 
-Before we look at the spatial resampling methods implemented in `r mlr3spatiotempcv` we will first show what can go wrong if standard resampling methods are used.
-Below we benchmark a decision tree on the `r ref("mlr_tasks_ecuador")` task using two different repeated cross-validation resampling methods, the first ("NSpCV") is a standard resampling method from `r mlr3`, the second ("SpCV") is from `r mlr3spatiotempcv` and is optimized for spatial data.
+Before we look at the spatial resampling methods implemented in `r mlr3spatiotempcv` we will first show what can go wrong if non-spatial resampling methods are used for spatial data.
+Below we benchmark a decision tree on the `r ref("mlr_tasks_ecuador")` task using two different repeated cross-validation resampling methods, the first ("NSpCV" (non-spatial cross-validation)) is a standard resampling method from `r mlr3`, the second ("SpCV" (spatial cross-validation)) is from `r mlr3spatiotempcv` and is optimized for spatial data.
```
pat-s, line 813:
```suggestion
Below we benchmark a decision tree on the `r ref("mlr_tasks_ecuador")` task using two different repeated cross-validation resampling methods, the first ("NSpCV" (non-spatial cross-validation)) is a non-spatial resampling method from `r mlr3`, the second ("SpCV" (spatial cross-validation)) is from `r mlr3spatiotempcv` and is optimized for spatial data.
```

book/special.qmd
```diff
@@ -964,15 +842,14 @@ autoplot(rsmp_sp, task, fold_id = c(1:3), size = 0.7) /
   scale_x_continuous(breaks = seq(-79.06, -79.08, -0.02))
 ```
 
-Now we have seen why spatial resampling methods matter we can take a look at what is implemented in `r mlr3spatiotempcv`.
-At the time of writing we have implemented 19 different resampling methods for spatial and spatiotemporal analysis.
-These are categorized into:
+Now we have seen why spatial resampling matter we can take a look at what is implemented in `r mlr3spatiotempcv`.
```
pat-s, line 845:
```suggestion
Now we have seen why spatial resampling matters we can take a look at what methods are available in `r mlr3spatiotempcv`.
```

book/special.qmd
```diff
@@ -1039,9 +916,7 @@ plot(pred, col = c("#440154FF", "#443A83FF", "#31688EFF",
...
-Probabilistic regression extend deterministic regression by making predictions over a full probability distribution and not just a single point.
-You can train and predict probabilistic regression models with `r mlr3` or more effectively in pipelines with `r mlr3proba`.
-Survival analysis, also available in `r mlr3proba`, is a special form of probabilistic regression when the outcome is censored, which means it may never be observed within a given time frame.
+Survival analysis, available in `r mlr3proba`, can be thought of as a regression problem when the outcome may be censored, which means it may never be observed within a given time frame.
 The final task in `r mlr3proba` is density estimation, the unsupervised task concerned with estimating univariate probability distributions.
 Using `r mlr3cluster`, you can perform cluster analysis on observations, which involves grouping observations together according to similarities in their variables.
 Finally, with `r mlr3spatial` and `mlr3spatiotempcv`, it is possible to perform spatial analysis to make predictions using co-ordinates as features and to make spatial predictions.
```
pat-s, line 922:
```suggestion
Finally, with `r mlr3spatial` and `r mlr3spatiotempcv`, it is possible to perform spatial analysis to make predictions using co-ordinates as features and to make spatial predictions.
```

pat-s (changes requested):
Thanks for the implementation of changes! Definitely an improvement over the previous one and well appreciated. Even though I struggled a bit understanding the changes made and the thoughts behind as these were not shared alongside the changes :) I didn't recheck all my previous comments again but I am almost happy with the current state (see below) 🙂:

Additional improvement suggestions which are not in the diff:

> In the example below we load the leipzig_points file for training

-> "In the example below we load the leipzig_points points dataset for training"

> Both files are shipped with, and can be downloaded from, [mlr3spatial](https://mlr3spatial.mlr-org.com/).

-> Both datasets are included as examples data within [mlr3spatial](https://mlr3spatial.mlr-org.com/).

## Spatial prediction

Here we're currently still missing mentioning the actual motivation for the existence of {mlr3spatial}, i.e. what it does to make the lifer easier for users:

"This seemed easy, right? Under the hood, mlr3spatial takes away a lot of cumbersome handling and coercing between R objects and formats - including taking care to not loose the spatial reference of the data.
If you would attempt this from scratch yourself, you would need to coerce between spatial objects and plain data.tables a lot. Not even mentioning handling the prediction step in a way that your memory is not overflowing but still allowing you to use more than one care of your machine.
All of this is handled by mlr3spatial to make the modeler's life easier. Depending on the data and your machine, it might also outperform the native parallelization methods of the spatial packages as shown [in this benchmark conducted by us](https://mlr3spatial.mlr-org.com/articles/benchmark.html).

*Feel free to rephrase above as needed. I guess Marc could also have a look at this as he implemented most of the logic in mlr3spatial.*

PR #563, opened 2023-04-03
[WIP] Chapter 8.n Fairness

pfistfl:
This is currently the last sub-chapter in 8-special. 

**Open Questions**
- [x] Which chapter does this belong to?
- [x] Does the current depth work? Do I need to shorten further?


larskotthoff:
- Suggest to put the submitted paper referenced at the beginning on arxiv so it can be referenced properly. A reference to a paper that isn't available isn't helpful for the reader.
- 10.1.1.1 "In this article" -> "In this chapter"
- 10.1.1.3 there's a disconnect between the example and the code. How does the measure know what groups to compare? Is it always for two groups, and if not, how are differences for more groups aggregated? Why does it go to -\infty if it's a difference? Or is it not?
- 10.1.1.4 example should be moved one level up in the section hierarchy.
- 10.1.1.6 method of constructing a fairness measure from a string "fairness.<base>" isn't explicitly mentioned before.
- 10.1.2 `po("EOd")` should be mentioned in the text as it's not obvious what it does.
- 10.1.2 results for the example (table or figure?) don't show up for me. Suggest using `fairness.acc` as measure to be consistent with the example before.
- 10.1.2 would be good to have an example for training a fair `fairzlrm` model.
- 10.1.2 would be good to have a small example for tuning fairness with pipelines at the end. Reference to a paper that is not available for such an example is particularly problematic.
- 10.1.3 documentation code example doesn't show how to specify a folder.
- 10.1.3 visualization should come much earlier, i.e. in the section on the example dataset.
- 10.1.3 visualization code example gives warnings and is not consistently formatted. Unclear why train/test sets are filtered. Text says decision tree and random forest are trained, but only random forest is trained.
- No exercises or resources at the end of the chapter.

Overall, this chapter feels much more like a technical paper than a book chapter. I suggest starting with an example that shows unfairness/bias in some way and reduce the amount of background material. In particular, the notation in 10.1.1.3 is not explained and doesn't seem to be necessary to understand the concept.

berndbischl:
i have read @larskotthoff comments. they all sound plausible. the chap should not become toooo long.

theres one thing where i dont know whether thats too much.

"10.1.2 would be good to have a small example for tuning fairness with pipelines at the end. Reference to a paper that is not available for such an example is particularly problematic." 

i am fine if this cannot be done. if you have, flo, in a short form you can include it. but i would resolve this last

berndbischl:
and one other thing:

"Overall, this chapter feels much more like a technical paper than a book chapter. I suggest starting with an example that shows unfairness/bias in some way and reduce the amount of background material. In particular, the notation in 10.1.1.3 is not explained and doesn't seem to be necessary to understand the concept."

i wouldnt be "that harsh". also, this is more a "researchy" topic -- at the end of the book. 
IMHO if your other issues are resolved, this criticsim isnt that much warranted. i think the intro with the background stuff is fine.

i would suggest to work on the conrete stuff from the bulletpoint list

berndbischl:
i guess showing an "unfair example" in the beginning WOULD be very nice though.... and to use it as a running example. at least where possible

pfistfl:
Thanks a lot, @larskotthoff 

I will iterate and take the feedback into account.

* The paper will be published in the next R Journal Volume. 
  I would hope that this is before the book is published(?)

larskotthoff:
Is the paper going to be published before the end of May?

If `fairzlrm` doesn't work easily, I wouldn't mention it. This seems like something a reader might try (and then be frustrated with).

I think exercises would be good here. Nothing big -- essentially the example on a different dataset, or evaluated in a slightly different way.

pfistfl:
I have added the exercises and solutions.
Paper: I do not have feedback yet. I will try uploading to arxiv tomorrow just in case.

larskotthoff:
Current version looks good to me.

RaphaelS1:
@pfistfl   are you happy for me to fix any conflicts, merge, and move into my column?

PR #572, opened 2023-04-14
address all comments in Chapter 8 Comments - closes #571

### LINKED ISSUE #571: Chapter 8 Comments
After reading through chapter 8 I had a few comments/questions/suggestions:

### Survival Analysis

- [x] In the second paragraph, there are many variables like $\Delta, Y, ...$ introduced. It would be helpful to the reader for these to be compared to the practical example from the previous paragraph like for example, $\Delta$ is the event indicator which in the previous example would indicate if the runner finished the race.
- [x] Right before the TaskSurv section there is a plot that illustrates examples of right censoring, it would be helpful if the plot was introduced right before. This should answer questions like what is this kind of plot called, what does it show? etc.
- [x] Before code chunk `special-011` where `as_task_surv()` is called please expand on the many parameters: what do they mean? what data types do they accept?
- [x] In the same code chunk please separate the `autoplot()` call and introduce the plot. Explain what type of plot it is and what it means/how to interpret it briefly.
- [x] Before code chunk `special-016` please add a sentence or note about the parameter `gamma.mu` and why it is being used here.
- [x] Similar to before the plot from code chunk `special-017` is lacking an explanation of what it is showing. Also a legend in the plot would be helpful.

### Density Estimation

- [x] Similar to before the plot from code chunk `special-024` is lacking an explanation of what it is showing. 
- [x] In section MeasureDens the measure logloss is introduced, my question after reading is: What does a good or bad logloss look like?

RaphaelS1:


nfoss2 (approved):
These changes look great!

PR #578, opened 2023-04-21
Chapter 2 RS rewrite

RaphaelS1:
@larskotthoff  @berndbischl  this is not finished and will need significant copyediting from me after but I've made big structural changes that I'd like you to look at before I do more work (hence leaving as a draft)

In summary:

1. Renamed chapter to be more specific
2. Slight rewrites of intro, mainly spelling/grammar 
3. Restructured Tasks, mainly reordering of content, removing some tipboxes, adding some clarifications and examples
4. For Learners mainly just spelling/grammar rewrites to improve flow of the text, most significant difference is in the ordering. Hyperparameters now at the end, predict types closer to the beginning. Basically before there were lots of things being referenced out of order, that should all be fixed now. Before there were also lots of unneccessary suubsections by themselves (in general we should always avoid something like 2.1.1 if there's no 2.1.2).
5. For predicting and in other places, whenever there's text like "there are many ways to do this but we don't recommend this" I've just added the recommendation and removed others. If people want to explore methods not recommended then they can look at online documentation, the book should only promote best practice
6. Evaluation basically same but reordered and added subsections so it's easier to follow
7. Classification had significant duplication, I've removed all of it. Now it summarizes everything in an experiment, then just highlights key differences - I am least happy with this after my edits and it still needs more work - expect typos and unfinished sentences when you read it
8. I haven't touched Column Roles. It definitely doesn't make sense where it is - we need to discuss this
9. I moved supported algorithms to the end
10. Added a conclusion and edited one of the exercises


larskotthoff:
Comments so far:
- Chapter rename seems redundant -- the book is on mlr3 and we don't cover fundamentals of ML or anything like that.
- Dashes seem to be inconsistent (- vs. --).
- Figure 2.1 not sure how useful this really is. We have figures illustrating specific concepts later, so I would be tempted to remove it.
- "There are many ways to evaluate models" reference looks wrong -- should that be to the evaluation chapter?
- 2.2 "decision tree regression learner" sounds weird -- regression tree learner
- 2.2.1.1 talks about the default split for partitioning in the text, but sets custom values in the code.
- 2.2.1.1 and 2.2.2 mentions the `row_ids` argument of train and predict, but passes row IDs as a unnamed argument.
- 2.2.2 "The final step of the basic ML workflow (Figure 3.1)" references figure in next chapter.
- 2.2.3.1 has a stray sentence at the end.
- 2.2.3.2 example doesn't output `maxdepth`.
- 2.3.4 wrong measure used in second code example.
- 2.5.2 not sure if the example for "misleading" positive/negative classes is useful -- could just refer to the thresholding section.
- 2.5.4 text describing confusion matrix and confusion matrix don't agree.
- 2.5.4 Thresholding -- I see the point of not repeating the same code, but your code example to compare the different thresholds in one plot is now much more complex. Not sure if this is suitable for the "Fundamentals" chapter. The "after" is actually worse here.

- Not sure about the removal of the example that showed the effect of changed hyperparameters (the different regression tree). I think it's useful to show this, instead of just saying that HPs matter.
- I like the regression/classification experiment sections.

RaphaelS1:
@larskotthoff 

>  Figure 2.1 not sure how useful this really is. We have figures illustrating specific concepts later, so I would be tempted to remove it.

Just to confirm do you mean this one? I can't remember if I added this by copying from c3 or we always had two copies? I think it's useful to highlight early on the high-level procecedure

Screenshot 2023-05-09 at 10 30 22

>  2.2.1.1 talks about the default split for partitioning in the text, but sets custom values in the code.

I think I did this as a deliberate choice just to show users how to do this and so they know what the default is if they don't change the parameter. Do you think it's confusing for the reader?

> 2.2.3.2 example doesn't output maxdepth.

I think this is a cache issue; will look into it

>  2.5.2 not sure if the example for "misleading" positive/negative classes is useful -- could just refer to the thresholding section.

I think this is not quite the same as thresholding as the example here is just showing how mlr assumes the positive class. But happy to remove if you think it isn't necessary

>  2.5.4 text describing confusion matrix and confusion matrix don't agree.

Which text?

RaphaelS1:
@larskotthoff 

>  Figure 2.1 not sure how useful this really is. We have figures illustrating specific concepts later, so I would be tempted to remove it.

Just to confirm do you mean this one? I can't remember if I added this by copying from c3 or we always had two copies? I think it's useful to highlight early on the high-level procecedure

Screenshot 2023-05-09 at 10 30 22

>  2.2.1.1 talks about the default split for partitioning in the text, but sets custom values in the code.

I think I did this as a deliberate choice just to show users how to do this and so they know what the default is if they don't change the parameter. Do you think it's confusing for the reader?

> 2.2.3.2 example doesn't output maxdepth.

I think this is a cache issue; will look into it

>  2.5.2 not sure if the example for "misleading" positive/negative classes is useful -- could just refer to the thresholding section.

I think this is not quite the same as thresholding as the example here is just showing how mlr assumes the positive class. But happy to remove if you think it isn't necessary

>  2.5.4 text describing confusion matrix and confusion matrix don't agree.

Which text?

PR #579, opened 2023-04-21
Optimization v5

be-marc:
## Lars comments

- [x] Probably not necessary to have the code to set the seed show up in the book. If we keep it, there should be some explanation. @RaphaelS1 
- [x] Would mention at the beginning that parameters determine the model and hyperparameters determine the way the model is built to make this connection between them more explicit. @sumny 
- [x] Would also mention at the beginning that methods like BO are adaptive, i.e. later iterations depend on what happened in earlier ones. @sumny 
- [ ] Figure 4.1 looks weird in the PDF, e.g. the i in Configuration is missing. Something gone wrong with the conversion?
- [ ] Colors in Figure 4.1. aren't clear. I think blue and purple are swapped? Some objects belong to different categories, which makes it more unclear. I would just highlight the user-supplied objects in the figure with a different color. @Coorsaa 
- [x] 4.1.1 in the example at the end for constructing a learner with `to_tune()` it might be worth pointing out that this learner will give an error message when trained as a "normal" learner.
- [x] 4.1.2 explain the meaning of the `threshold` argument for stagnation.
- [ ] I would move 4.1.4 Tuner immediately after 4.1.1 Learner and Search Space. Feels weird about terminating without knowing yet what will be terminated, especially with references to terminator none for grid search.
- [x] Code example in 4.1.3 `resampling` and `measure` aren't used.
- [x] Table 4.2 shows all packages as mlr3*, which seems to suggest that all tuners are implemented by the mlr3 team. Would make more sense to list the underlying package there, and mlr3tuning only if the particular tuner is actually implemented in there.
- [x] Table 4.2 caption mentions "methodology", but the table doesn't seem to contain that.
- [x] 4.1.4 paragraph about grid search doesn't explain how values for categorical parameters are chosen if the resolution is less than the total number of values. @sumny 
- [x] 4.1.4 paragraph on grid search I would replace "naive" with "non-adaptive". @sumny 
- [x] 4.1.4 in the paragraph on BO I would mention that the acquisition function also (potentially) uses the uncertainty of predictions of the surrogate model. @sumny 
- [x] 4.1.4 in the paragraph on multi-fidelity HPO I would make the connection to race where evaluation on subsets of folds is already mentioned (and maybe that's a better place to talk about racing). @sumny 
- [x] 4.1.4 paragraph on multi-fidelity HPO mentions anytime performance without explaining it. @sumny 
- [ ] 4.1.4 is there a way to determine automatically what tuner works on what space? Section talks about hierarchical spaces, non-numeric HPs etc which not all tuners can handle. Can a user find the list of "applicable" tuners automatically, similar to learners that can handle a task? If so, this should be mentioned, if not, could think about implementing this (but maybe not for this book). @sumny 
- [x] 4.1.4 paragraph on choosing the right tuner should briefly mention grid search with a coarse resolution and why that might not be a good idea. @sumny 
- [x] 4.1.4 same paragraph might be worth mentioning that if the budget is extremely small (e.g. only the number of initial samples for BO), random search usually does at least as well as adaptive methods. In particular suitable for quickly trying HPO. @sumny
- [ ] End of 4.1.4 mentions `batch_size` without details -- should at least mention that depending on the tuner values larger than 1 might not "work".
- [ ] ParamSet at the end of 4.1.4 should mention what `param_resolutions` means.
- [x] 4.1.5 difference between `x_domain` and `learner_param_vals` is unclear -- sounds like `learner_param_vals` is a strict superset? Then why store both? Might be helpful to show both.
- [x] 4.1.6 mention briefly why a log scale makes more sense for these hyperparameters than a linear scale. At least point out for the example that we got a much better result than when tuning on a linear scale before.
- [x] 4.1.7 text says with `tune()` I have less control, but code seems to do exactly the same thing. Give example where I can't use `tune()`.
- [x] 4.1.8 would say a little bit more about how the plot is generated, in particular that this requires training a surrogate model (which learner is used for that?).
- [x] 4.1.8. might be worth adding an example that uses the same space, but BO and then plot the result to show that it focuses on the area where the lowest error is.
- [x] 4.2 can I get the archive from an autotuner and analyze the tuning result in the same way as shown before?
- [ ] 4.3 would be good to have a little bit more explanation at the beginning, e.g. the usual thought experiment about HPs that don't do anything but there's random noise in the evaluation.
- [x] Figure 4.4 colors in the figure and caption are different.
- [x] 4.3 at the end of the description of nested resampling it might be worth mentioning explicitly what the "output" of each level is -- inner resampling produces generalization performance estimates based on which a single HP configuration is chosen to be evaluated on the outer resampling; outer resampling produces generalization estimates for these optimal HP configs. In particular the outer resampling does not produce an optimized HP config. Might also be helpful to say how many times total HPO will be run in this example.
- [x] 4.3.1 at the end say explicitly how HP config to fit a final model should be obtained.
- [x] 4.3.2 mentions overfitting without explaining it.
- [ ] 4.3.2 is unclear to me -- inner performances 0.23 0.15 0.19 don't look significantly better than outer performances 0.17 0.17 0.23. Should also point out that the inner resampling results are aggregated over the inner folds while the outer ones are individual performance numbers. Text then says that significant difference indicates overfitting, but when aggregating the outer fold performance estimates we somehow get an unbiased estimate?
- [x] 4.3.3 "Instructive Example" -> "Detailed Example"?
- [x] 4.3.3 introduces data generators and `tgen()` without explanation. Briefly describe.
- [x] 4.3.3 none of the code examples produce output, but it looks like some of them should.
- [ ] Unsure about 4.3.3 in general. Nested resampling only makes a very small difference. Maybe kNN and tuning k would show a larger difference? At the moment I'm not sure if this example adds anything useful -- the previous section already shows that there are (small) differences in the performance estimates. An additional example would need to show a much larger difference.
- [ ] 4.4.1 in the first code example say explicitly that "evaluate" arguments refer to the evaluate package. What are other possible arguments, e.g. if I want to use callr?
- [x] 4.4.1 mention why one might want to use callr -- the text makes it sound like there's no advantage as it adds additional overhead.
- [x] 4.4.1 what happens if I don't set a fallback learner?
- [ ] 4.4.1 would be good to show an example where an error actually happens. Maybe subset sonar to contain only one class? That might cause some errors? Or set a really low timeout.
- [x] 4.5.1 what happens when a search space is not bounded? Error message? Infinite loop?
- [ ] 4.5.1 for `ParamUty` it might be worth pointing out that while this cannot be in the *search space* paramset, it can be in the paramset of a learner to be tuned.
- [x] 4.5.1 don't really understand the example with the "numeric" factors (and the list of values is different in different code snippets) -- is the equivalent code with the trafo function generated automatically or how does the coercion to/from characters happen? So is the `p_fct(c(0.3, 0.7))` essentially syntactic sugar?
- [ ] 4.5.2 should this be labelled advanced as well?
- [ ] 4.5.2 don't understand the initial example. Why can I not just say `depends = cost == 0.3`? Why do I need the designation as a special named value? The output says "Trafo is set." but the code doesn't set a trafo. @sebffischer 
- [ ] 4.5.3 says tuning spaces are taken from publications that are encoded in the name, but doesn't list the publications -- where do "default" and "rbv2" spaces come from? Is the list of publications somewhere in the documentation?
- [ ] 4.6.1 not sure if this section is necessary. We don't give that much detail on any of the other methods, and it's not clear that this level of details is beneficial for the reader. I would remove this section and simply point to the paper.
- [ ] 4.6.2 would be shorter to use `lts()` to retrieve the default space and modify only the `nrounds` HP.
- [x] 4.6.2 r_min and r_max are defined both when the HP space is created and when HB is instantiated. Are both necessary? Also should `hyperband_schedule()` be `tuner$hyperband_schedule()`? Or is this call not actually modifying anything but only illustrating what the schedule looks like? If so say that explicitly.
- [x] 4.6.2 "This does not necessarily have to be a configuration evaluated with the highest budget since XGBoost can in general overfit the data when using too many boosting iterations." doesn't make sense to me. Does this mean that XGBoost automatically detects overfitting and some how prevents HB from evaluating with the highest budget?
- [ ] 4.6.3 would swap that with 4.6.2 so that the "familiar" example of tuning the SVM comes first. Or actually maybe this should be moved to the pipelines chapter.
- [ ] 4.6.3 could we use `lts()` to get this search space and only modify the budget param?
- [x] 4.6.3 same as in 4.6.2 -- do the budget ranges need to be specified twice?
- [x] 4.6.3 text says that best model is SVM with radial kernel, but code output shows poly kernel.
- [x] Figure 4.9 looks like there's only a single blue dot? How can the number of selected features be fractional? I would move the reference to this figure to the end of the section and in the beginning only say that there will be a figure to illustrate Pareto front. @sumny
- [ ] 4.7 difference between Pareto set and Pareto front not entirely clear -- I would phrase this as the Pareto set defining the Pareto front, along which we can get the values for points not in the Pareto set or something like this. @sumny
- [x] 4.7 the best configurations shown at the end of the section have each exactly the same values for both objectives. Might be better to have an example where the result is more "interesting". @sumny
- [ ] Table 4.5 R6 class names protrude into the summary column in the PDF.
- [x] adjust pareto front @sumny


be-marc:
> End of 4.1.4 mentions batch_size without details -- should at least mention that depending on the tuner values larger than 1 might not "work".
 
I think our explanation is good enough.

https://github.com/mlr-org/mlr3book/blob/11805043af0936775f42b1a0e0fa90f9999faea8/book/optimization.qmd#L253

> ParamSet at the end of 4.1.4 should mention what param_resolutions means.

We don't have to explain the parameter because we don't use it.

sebffischer:
@larskotthoff this is explained in section 4.5.1 after "If, for example, only the values 0.1, 3, and 10 should be tried for the cost parameter, even when doing random search, then the following search space would achieve that:"

be-marc:
> @larskotthoff this is explained in section 4.5.1 after "If, for example, only the values 0.1, 3, and 10 should be tried for the cost parameter, even when doing random search, then the following search space would achieve that:"

Okay now I understand the example. I think the example is too artificial and might confuse the reader.  Why don't we just show the second example with `degree`? We say that the dependency was already set but do it again as an example. And then a note that setting dependencies in pipelines is very useful.

```r
learner = lrn("classif.svm",
  kernel = to_tune(c("linear", "polynomial")),
  degree = to_tune(p_int(2, 5, depends = kernel == "polynomial"))
)
``

**Comment by sebffischer** [comment]
> > @larskotthoff this is explained in section 4.5.1 after "If, for example, only the values 0.1, 3, and 10 should be tried for the cost parameter, even when doing random search, then the following search space would achieve that:"
> 
> Okay now I understand the example. I think the example is too artificial and might confuse the reader. Why don't we just show the second example with `degree`? We say that the dependency was already set but do it again as an example. And then a note that setting dependencies in pipelines is very useful.
> 
> ```r
> learner = lrn("classif.svm",
>   kernel = to_tune(c("linear", "polynomial")),
>   degree = to_tune(p_int(2, 5, depends = kernel == "polynomial"))
> )
> ```

yes feel free to change it I am now on holidays

File: book/optimization.qmd [hunk 1]
```diff
@@ -289,7 +287,9 @@ The column `x_domain` contains transformed values (see @sec-logarithmic-transfor
 
 ### Logarithmic Transformations {#sec-logarithmic-transformations}
 
-The SVM's `cost` and `gamma` parameters that we have tuned in the example above, are usually tuned on a ` r define("logarithmic scale")` (as opposed to linear).
+The SVM's `cost` and `gamma` parameters that we have tuned in the example above, are usually tuned on a `r define("logarithmic scale")` (as opposed to linear).
+The parameters vary over several orders of magnitude.
+Using a logarithmic scale allows us to efficiently search over a wide range of values.
```
be-marc, line 292: and focusing on lower values.

book/optimization.qmd
```diff
@@ -420,7 +421,7 @@ Although we evaluated 25 configurations in total, only 5 unique hyperparameter v
 
 ```{r optimization-019}
 #| label: fig-surface
-#| fig-cap: Model performance with different configurations for `cost` and `gamma`. Bright yellow regions represent the model performing worse and dark blue performing better. We can see that high `cost` values and low `gamma` values achieve the best performance. Note that we should not directly infer the performance of new unseen values from the heatmap since it is only an interpolation based on a surrogate model. However, we can see the general interaction between the hyperparameters.
+#| fig-cap: Model performance with different configurations for `cost` and `gamma`. Bright yellow regions represent the model performing worse and dark blue performing better. We can see that high `cost` values and low `gamma` values achieve the best performance. Note that we should not directly infer the performance of new unseen values from the heatmap since it is only an interpolation based on a surrogate model (`classif.ranger`). However, we can see the general interaction between the hyperparameters.
```
be-marc, line 424: `regr.ranger`

book/optimization.qmd
```diff
@@ -757,6 +775,10 @@ instance = tune(
 as.data.table(instance$archive)[, .(cost, gamma, classif.ce, errors, warnings)]
 ```
 
+::: {.callout-note}
+When tuning, encapsulation should always be combined with a fallback learner, because the archive must contain a performance value for each configuration.
```
be-marc, line 779: What fallback learner?

larskotthoff:
> > End of 4.1.4 mentions batch_size without details -- should at least mention that depending on the tuner values larger than 1 might not "work".
> 
> I think our explanation is good enough.

Ok, so let's say the reader sets `batch_size = 10` as in the example, but switches to BO as that's obviously a better choice. Then notices that the batch size is essentially 1 and gets confused. Adding a sentence that the effect of `batch_size` also depends on the HPO method, and that in particular it may be smaller than requested, would be helpful I think.

> > ParamSet at the end of 4.1.4 should mention what param_resolutions means.
> 
> We don't have to explain the parameter because we don't use it.

But it's listed in the output. A very short explanation would be sufficient, but having this show up in the output and not mention it at all would not be good I think.

> > 4.3.2 is unclear to me -- inner performances 0.23 0.15 0.19 don't look significantly better than outer performances 0.17 0.17 0.23. Should also point out that the inner resampling results are aggregated over the inner folds while the outer ones are individual performance numbers. Text then says that significant difference indicates overfitting, but when aggregating the outer fold performance estimates we somehow get an unbiased estimate?
> 
> We have already tried different splits. The result remains the same. We cannot change it.

But the sentence after the example could be changed to "Significantly lower predictive performances on the outer resampling *would* indicate that the models with the optimized hyperparameters overfit the data."

> > 4.4.1 in the first code example say explicitly that "evaluate" arguments refer to the evaluate package. What are other possible arguments, e.g. if I want to use callr?
> 
> I think this is properly explained in the third sentence after the code chunk.

The sentence only mentions the packages. Are the valid arguments "evaluate" and "callr"? Then mention that. Otherwise the reader has to infer that the arguments have the same name as the underlying packages.

> > 4.5.3 says tuning spaces are taken from publications that are encoded in the name, but doesn't list the publications -- where do "default" and "rbv2" spaces come from? Is the list of publications somewhere in the documentation?
> 
> Yes the names are badly chosen. We should discuss this again.

And for the book add a sentence on what the names mean?

be-marc:
> Ok, so let's say the reader sets batch_size = 10 as in the example, but switches to BO as that's obviously a better choice. Then notices that the batch size is essentially 1 and gets confused. Adding a sentence that the effect of batch_size also depends on the HPO method, and that in particular it may be smaller than requested, would be helpful I think.

Added:

The `batch_size` parameter is only available for `tnr("random_search")` and `tnr("grid_search")` and is set to 1 by default.
The other tuners set the batch size themselves.

> But it's listed in the output. A very short explanation would be sufficient, but having this show up in the output and not mention it at all would not be good I think.

Bernd told us to not explain it because we don't use it. 

> But the sentence after the example could be changed to "Significantly lower predictive performances on the outer resampling would indicate that the models with the optimized hyperparameters overfit the data."

:thumbsup:

> The sentence only mentions the packages. Are the valid arguments "evaluate" and "callr"? Then mention that. Otherwise the reader has to infer that the arguments have the same name as the underlying packages.

Added:

The possible options are `evaluate` and `callr` named after the packages of the same name.

> And for the book add a sentence on what the names mean?

Okay `{learner-id}.{publication} was misleading. Changed to

The tuning spaces are named according to the scheme `{learner-id}.{tuning-space-id}`.
The `default` tuning spaces are published in @hpo_practical.
More tuning spaces are part of the random bot experiments `rbv1` and `rbv2` that are published in @kuehn_2018 and @binder2020.

larskotthoff (approved):

larskotthoff:
Thanks, looks good.

PR #587, opened 2023-05-05
Chapter5 rsreview

RaphaelS1:
As with my other reviews, will need copyediting afterwards to fix spelling and grammar issues (mainly added by me in this review). @larskotthoff @berndbischl - ready for you to review. @mnwright also pinging you so you're aware of changes (and can comment on them if you want).

Overall thoughts: chapter read very well, easy to read and understand. However there was significant duplication in the Wrappers section with Chapter 4 which I have now removed and shortened the chapter considerably. I've also changed the initial examples from classif to regr to be more consistent with the rest of the book. I have also added in some non-mlr examples near the beginning to help motivate filters a bit more. Finally I've edited the exercises slightly to be more in line with other exercises in the book.

@mnwright the changes will look more significant than they are in the diff because there's a lot of cutting and pasting to move content around.


mnwright:
What about #558?

Also, you re-wrote and re-structered the whole chapter after we had several rounds of reviews with @berndbischl and @larskotthoff and everyone was happy. Now someone has to review your new chapter again. 

To be clear, I don't agree with many of the changes and suggest to go back to the version in #558.

larskotthoff:
I would defer to @mnwright here and revert back to the previous version.

RaphaelS1:
Hi @mnwright , the review cycle we have setup is that something is in my column if it's been merged into main - no one told me about #558  -- clearly I've wasted your time *and* my time. I've merged 558 and will review next week.

RaphaelS1:
Not your fault! And we're going to keep it as-is. I was a bit concerned originally about the duplication between HPO and this chapter (hence removing part of the wrappers section), this would be especially 'bad' in the current book structure, However we are proposing a book restructure to the publisher that will actually require more detail anyway as the chapters won't follow one just after the other.

PR #591, opened 2023-05-05
Chapter [?] Large-Scale Benchmark Experiments (v3)

sebffischer:


larskotthoff:
- [ ] The definition of each experiment doesn't include the measure, which is included in the table. I think it would be more intuitive to define this with the measure -- I understand the motivation that to compute different measures, the experiment doesn't have to be rerun, but conceptually it makes more sense to make the measure part of the experiment I think.
- [ ] Notation inconsistent in the intro; k is the number of tasks and resamplings and also the running index for measures.
- [ ] The chapter assumes that there's a 1-1 correspondence between tasks and resamplings -- with this, I think it would make more sense to define this as one entity (resampled task?) rather than two. Unless we want to permit different resamplings for each task (which I think would also make sense).
- [ ] Not sure about Table 9.1 -- I don't think it makes the concept that much clearer. Maybe a graph instead? But this is also already explained in chapter 3. I think the table can be removed.
- [ ] First code example: I would explain how exactly `robustify()` works here (this isn't explained in detail anywhere else in the book as far as I can see). In particular, it seems strange to have the same learner appear twice.
- [ ] 9.1.1 is there a way to force the download of the entire OML dataset? E.g. if I want to work offline.
- [ ] 9.1.2 might be worth pointing out that the task here is different to the mlr3 task that was created just before.
- [ ] 9.1.4 explain how `paired = TRUE` works for `benchmark_grid()` -- does this only consider tasks and resamplings or learners as well, what happens if tasks and resamplings are not exactly the same length, etc.
- [ ] 9.2 mention at the beginning that it's usually a good idea to run a small version of the benchmark experiment first, in particular to estimate how long the full experiment will take. The note mentions recommendations for experiments that complete within a couple of hours, but how do I know that?
- [ ] 9.2.2 not sure if it's necessary to introduce additional notation for experiments; it doesn't seem to be used anywhere later.
- [ ] 9.2.2 don't understand why this section is called "two worlds". I think this section and 9.2.3 can be combined.
- [ ] 9.2.3 first code output doesn't show the code that was run to obtain it.
- [ ] 9.2.4 why is it more efficient to have one parameterized algorithm than multiple?
- [ ] 9.2.5 I would show the output of `testJob()` here.
- [ ] 9.2.5 what exactly does the slurm-simple template define?
- [ ] 9.2.5 when is it necessary and when not to save the registry?
- [ ] 9.2.5 in the chunking IDs example, where does the 36 come from?
- [ ] 9.2.5 where I can find a list of possible resources to specify? E.g. if I need a GPU.
- [ ] 9.2.5 tip on persistent R sessions: mention that this is not necessary and may not be possible (e.g. the head node kills all user processes when the user logs out) and clarify "multiple SSH sessions" -- I think this just means running things in the background after logging out?
- [ ] 9.2.5/9.2.6 add something that shows how to load an existing registry (e.g. if there's no persistent R session), show an example for `loadResult()`, and move the last part of 9.2.5 to 9.2.6 so that there's a logical progression job submission -> job monitoring -> job evaluation. I would rename 9.2.6 "Job Monitoring, Evaluation, and Error Handling" or something like that.
- [ ] 9.2.6 I would simply analyze the warnings from the actual `large_design` here instead of constructing a new one with the debug learner.
- [ ] 9.4. overview table: I would remove functions that have been used elsewhere and are not specific to the topic: autoplot(), as_benchmark_aggr().

sebffischer:
TODO for myself: 

document 
```
Sys.setenv(OMP_NUM_THREADS=1)
Sys.setenv(OMP_THREAD_LIMIT=1)
```

RaphaelS1:
@sebffischer  are you happy for me to fix any conflicts, merge, and move into my column?

larskotthoff (approved):

sebffischer:
sorry i still have some small edits that i will push tomorrow will message you then

sebffischer:
This PR: https://github.com/mlr-org/mlr3book/pull/600 is essentially the same except for an update package and a change of the caching mechanism :)

PR #593, opened 2023-05-11
Use as_benchmark_aggr

olivroy:
Warning: 'as.BenchmarkAggr' is deprecated.
Use 'as_benchmark_aggr' instead.
See help("Deprecated")


RaphaelS1:
Thanks ! I've had this on my todo list for ages

PR #603, opened 2023-05-19
streamlining BO

sumny:
In progress, I'm working on streamlining the  (BO) chapter to enhance accessibility and improve clarity for easier comprehension.


sumny:
- [ ] Don't emphasize surrogateLearner vs SurrogateLearnerCollection
- [ ] Only mention the different Obejctive classes and their differences in a callout box
- [ ] Removed SMS ego as an example of another MO BO Loop
- [ ] Reduce math and notation to the minimum level

sumny:
After merge open issue with the following todos prior to hand in:

- [ ] CRAN release of new mlr3mbo and using that version in renv
- [ ] include new figure made by stefan

PR #620, opened 2023-06-06
Preprocessing: Exercise and Solution

ja-thomas:


RaphaelS1:
Thanks @ja-thomas  will merge when passes

RaphaelS1 (approved):

sumny (approved):
In general I really like the exercise and I believe readers learn a lot!
I have some few minor comments that you might want to consider before merging:

* not sure if it is clear that one has to load `mlr3data` to get access to the `kc_housing` task
* dropping the `date` feature and then readding later could also be done via the `$col_roles` AB
  `kc_housing$col_roles$feature = setdiff(kc_housing$feature_names, "date")`
  and then later
  `kc_housing$col_roles$feature = c(kc_housing$feature_names, "date")`
* conversion to a factor can also be done more efficiently via `po("colapply")` in the case of more than one feature that should be converted
  `po("colapply", applicator = as.factor, selector = select_name("zipcode"))`
* `ppl("robustify")` can also be given the task and learner during construction resulting in a more tailored pipeline
* maybe add a comment why we instantitate the resampling (so that we use the same for the different pipelines)
* not really clear where the featureless learner comes from (ofc it is meaningful to include it) but it is not stated in the exercise
* instead of manually converting `date` to a numeric value, we could also show `po("datefeatures")` or maybe state that this PipeOp also exists (in the case of wanting to use more advanced time and date features although I guess simply using year is the best choice here)
* any clue/comments why ` 7: regr.glmnet_preproc kc_housing_feat_extr 100228.45` performs worse than ` 9: regr.glmnet_robustify kc_housing_feat_extr 95457.60`?
* also might be useful to add one sentence why using the year and the distance to downtown might be a good idea

RaphaelS1:
@ja-thomas what's the status here? We need to merge ASAP

ja-thomas:
I can fix the comments on Monday, then it should be good to go

RaphaelS1:
@berndbischl please note this PR when you look at Exercises on Sunday

mb706:
is merged in #707

PR #622, opened 2023-06-06
Preprocessing:  Ames Data set

ja-thomas:
Add dataset for preprocessing chapter and code to (re-)create the dataset


ja-thomas:
Ok, I think this needs to be done in two PRs to pass since I have to reference this repo's main branch. 

This PR just adds the data folder keeping the URLs pointing at the original repo, when this is merged we can change the URLs. 

(If somebody has a better way, feel free)

PR #638, opened 2023-06-13
Fix author abbreviations for authors with more than one first name

mnwright:
My name was abbreviated with "N M" because of the middle initial.


RaphaelS1:
Good catch, sorry I didn't spot before

PR #647, opened 2023-06-16
Fix lsb

RaphaelS1:


book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -122,8 +122,13 @@ if (file.exists(path_odatasets)) {
...
 ```
 
+```{r}
+print(odatasets[, c("data_id", "name")])
+```
+
+
 We can see that some datasets have duplicated names, which is why each dataset also has a unique ID.
-By example, let us consider the 'adult' dataset with ID 1590.
+For example, let us consider the 'adult' dataset with ID 1590.
```
RaphaelS1, line 131: It is "by example", we are demonstrating the thing by using an example

book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -110,7 +110,7 @@ odatasets = list_oml_data(
   number_features = c(10, 20), # 10-20 features
   number_instances = c(45000, 50000), # 40,000-50,000 rows
   number_classes = 2 # 2 classes
-)[, c("data_id", "name")]
+)
```
RaphaelS1, line 113: Does removing this cause too much printout?

book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -139,7 +144,7 @@ odata$license
 c(nrow = odata$nrow, ncol = odata$ncol)
 ```
 
-If we want to access the actual data, then calling `$data` will download the data, import it into R, and then store the `data.frame` in the `OMLData` object:
+If we want to access the actual data, then accessing `$data` will download the data, import it into R, and then store the `data.frame` in the `OMLData` object:
```
RaphaelS1, line 147: Is `accessing` right? We're saying "to access data then access `$data`", reads a bit strangely. If you don't like calling then what about:

```suggestion
If we want to access the actual data, then the `$data` field will download the data, import it into R, and then store the `data.frame` in the `OMLData` object:
```

book/chapters/chapter11/large-scale_benchmarking.qmd
RaphaelS1:
```suggestion
By example, let us consider the 'adult' dataset with ID 1590.
```

book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -181,7 +186,7 @@ path_adult_tasks = file.path("openml", "manual", "adult_tasks.rds")
 
 ```{r, eval = !file.exists(path_adult_tasks)}
 # first 5 tasks making use of the adult data
-adult_tasks = list_oml_tasks(data_id = 1590, limit = 5)
+adult_tasks = list_oml_tasks(data_id = 1590)
```
RaphaelS1, line 189: We are deliberately suppressing a lot of output to save space in the book.

```suggestion
adult_tasks = list_oml_tasks(data_id = 1590, limit = 5)
```

book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -439,7 +434,7 @@ In this output we can see how each job is now assigned a unique `job.id` and tha
 
 With the experiments defined we can now submit them to the cluster.
 However, it is best practice to first test each algorithm individually using `r ref("batchtools::testJob()", aside = TRUE)`.
-By example we will only test the first job (`id=1`) and will use an external R session(`external = TRUE`).
+For example, we will only test the first job (`id=1`) and will use an external R session(`external = TRUE`).
```
RaphaelS1, line 446:
```suggestion
By example, we will only test the first job (`id=1`) and will use an external R session(`external = TRUE`).
```

book/chapters/chapter11/large-scale_benchmarking.qmd
mllg: this did not produce output (due to a bug in mlr3oml which returns the table invisible).

Instead, the subsetted table is now explicitly printed a few lines later.

book/chapters/chapter11/large-scale_benchmarking.qmd
mllg: This was super confusing because the "randomly picked id" was not part of the returned sample of 5 data sets. The output is still quite concise as only the ids are printed, not the table.

mllg:
Note that I've re-added a PNG which was missing - would it help to provide an SVG here instead?

book/chapters/chapter11/large-scale_benchmarking.qmd
RaphaelS1: Okay thanks

book/chapters/chapter11/large-scale_benchmarking.qmd
```diff
@@ -795,21 +625,23 @@ for (i in seq_along(tasks)) {
...
-
-The next step is to register the algorithm we want to run, which we achieve by calling `r ref("batchtools::addAlgorithm()")`.
-Besides the registry, it takes in the arguments:
+```{r large_benchmarking-051}
+#| label: fig-batchtools-illustration
+#| fig-cap: "Illustration of a batchtools problem, algorithm, and experiment. "
+#| fig-align: "center"
+#| fig-alt: "The diagram shows a rectangle that says 'static problem part, data', with an arrow pointing to 'dynamic problem function, fun(data, ...)' and 'algorithm function, fun(data, instance, ...)'. A box that says 'problem design, (addExperiments)' also has an arrow to the 'dynamic...' box. The 'dynamic...' box then has an arrow with text 'instance' that points to the 'algorithm function' box. A box that says 'algorithm design, (addExperiments)' also points to the 'algorithm function' box. Finally the 'algorithm function' box points to 'result'."
+#| echo: false
+knitr::include_graphics("Figures/tikz_prob_algo_simple.png")
```
RaphaelS1, line 634:
```suggestion
include_multi_graphics("tikz_prob_algo_simple")
```

this function auto choose png for PDF and svg for HTML

sebffischer:
Why is this: https://github.com/mlr-org/mlr3book/blob/6196d26d3e27126cce90be205e39f6440af9c355/book/chapters/chapter11/large-scale_benchmarking.qmd#L449 now commented out?

sebffischer:
Should we remove all "ID" with id? It is currently inconsistent. The rest of the book uses id I guess?

RaphaelS1:
@mllg @sebffischer can this be merged?

PR #664, opened 2023-06-20
fix: mbo exercise

sumny:


book/chapters/chapter5/advanced_tuning_methods_and_black_box_optimization.qmd
```diff
@@ -593,6 +593,8 @@ In `mlr3mbo`, the `r ref("SurrogateLearner")` is a higher-level R6 class inherit
 
 Any regression learner in `mlr3` can be used, however, most acquisition functions depend on both mean and standard deviation predictions from the surrogate model and therefore not all learners are suitable for all scenarios.
 Typical choices of regression learners used as surrogate models include `r index('Gaussian processes', "Gaussian process", lower = FALSE)` (`lrn("regr.km")`) for low dimensional numeric search spaces and `r index('random forests', 'random forest')` (e.g., `lrn("regr.ranger")`) for higher dimensional mixed (and/or hierarchical) search spaces.
+Note that `r ref("default_gp()")` and `r ref("default_rf()")` also grant convenient access to pre-configured Gaussian process and random forest learners, respectively.
```
RaphaelS1, line 596: Why does this refer to? Do these create `mlr3` objects? `Learners`?

book/chapters/appendices/solutions.qmd
```diff
@@ -363,8 +361,7 @@ instance = OptimInstanceMultiCrit$new(
 Using the surrogate, acquisition function and acquisition function optimizer that are provided, we can proceed to optimize the instance via ParEGO:
 
 ```{r}
-surrogate = srlrn(lrn("regr.ranger", num.trees = 10L, keep.inbag = TRUE,
-  se.method = "jack"))
+surrogate = srlrn(default_rf())
```
RaphaelS1, line 364: I don't think adding these functions here makes sense as you only have the small note below.  Either we need to expand upon the note more or go back to what we had before

sumny: Only included this to show where it goes but BB told me that exercises and solutions will be rewritten by him and Martin.

book/chapters/chapter5/advanced_tuning_methods_and_black_box_optimization.qmd
sumny: BB wanted these to have so that users do not always have to do something like
```r
surrogate = srlrn(lrn("regr.ranger", num.trees = 10L, keep.inbag = TRUE,
  se.method = "jack"))
```

in the exercises.
Instead they now can do

```r
srlrn(default_rf())
```

and to answer your questions: yes, they simply construct learners but set some hyperparameters to some specific values that we believe are suitable if the learner is going to be used as a surrogate learner in BO

RaphaelS1: What package are these functions from, I can't find them in mlr3verse or mlr3mbo?

sumny: they are now included in the new mlr3mbo version as suggested by BB (not yet on cran though)

PR #671, opened 2023-06-21
readd old anytime BO figure in case it will be used again

sumny:
readds the anytime BO figure in case it shall be included in the final version (as to be discussed and requested by BB).
@mb706


RaphaelS1:
@sumny I'm confused by this PR. You say re-add in case it will be used but then add the figure and text?

sumny:
@RaphaelS1 ah yes sorry, this PR should not be merged for now but is only there in case BB decides that the figure needs to be readded due to the exercises missing an example otherwise.

PR #709, opened 2023-07-02
Bb chapter 8, 2nd ed.

mb706:


mb706:
I am done here

book/chapters/chapter8/non-sequential_pipelines_and_tuning.qmd
```diff
@@ -561,6 +561,17 @@ Tuning the `selection` hyperparameters can help determine which of the possible
...
+# instead of plotting, we make autoplot() save the plot so we can edit it afterwards
+# This is *not* the same as ggplot::last_plot(), but the result is easier to handle in a loop.
+plt_container = new.env()
+autoplot = function(...) {
+  # <<- doesn't seem to work on CI for some reason
+  plt_container$plt = ggplot2::autoplot(...)
+  invisible(NULL)
+}
+```
+
```
RaphaelS1, line 574: @mb706 this does not work as intended. Please check in the pdf output but this is quite broken...

Screenshot 2023-07-03 at 08 44 04

Also figures I previously fixed are now printing badly again, e.g. this is now a full page

Screenshot 2023-07-03 at 08 44 31

book/chapters/chapter8/non-sequential_pipelines_and_tuning.qmd
```diff
@@ -572,12 +583,44 @@ graph_learner$param_set$set_values(
...
+  et = element_text(angle = 45)
+  plt[[i]]$theme$axis.text.x = et
+  ### Remove the legends
+  plt[[i]]$theme$legend.position = "none"
+}
+autoplot = ggplot2::autoplot # reset to original function
+print(plt)
+```
+
+As we can see in the plot, the KNN-learner with `k` set to `r instance$result$classif.kknn.k` was selected, which performs best in combination with the Yeo-Johnson transform.
```
RaphaelS1, line 623:
```suggestion
As we can see, the KNN-learner with `k` set to `r instance$result$classif.kknn.k` was selected, which performs best in combination with the Yeo-Johnson transform.
```

RaphaelS1 (changes requested):

book/chapters/chapter8/non-sequential_pipelines_and_tuning.qmd
mb706: I am fixing the warning message, and apparently the autoplot() got lost when merging master. but are you sure the plot you are showing is not an artifact of how you are rendering a single file, or broken cache? This one looks fine for me

mb706: (or maybe an old mlr3pipelines version?)

mb706:
I added some arguments to suppress warnings, but I notice that plotting breaks when the cache is in a broken state. If the result does not look like this, you can try deleting the cache
image

PR #716, opened 2023-07-13
Update README

RaphaelS1:


RaphaelS1:
Yeah we could make something in the future, but TBH we haven't lost much here, mainly just finer details

PR #736, opened 2023-10-10
fix: use old renv version and upgrade mlr3misc version

sebffischer:
The "large-scale benchmarking" chapter always threw because of a change in renv that started to install packages with --with-keep.source this had two problems:
1. methods of R6 intances had a srcref object containing a promise with a large environment
2. function parameters (as in the robustify pipeline) were also quite large because from there one can also reach enviroments where a lot of srcrefs are kept

While this is inherently an issue that needs to be addressed in renv -- i.e. there should be an option to not keep the source -- we here fix it temporarily by relying on an older renv version and bumping the mlr3misc version where the problem is solved

https://github.com/rstudio/renv/issues/1713
https://github.com/mlr-org/mlr3misc/issues/88

### DIFF
--- a/book/renv.lock
+++ b/book/renv.lock
@@ -3470,7 +3470,6 @@
       "RemoteRef": "main",
       "RemoteSha": "d50c4bc45f2b33f1e594b611048f008bcd0508e4",
       "Remotes": "alan-turing-institute/distr6, binderh/CoxBoost, mlr-org/mlr3proba,\n    RaphaelS1/survivalmodels, xoopR/param6, xoopR/set6",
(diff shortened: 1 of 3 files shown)


book/renv.lock
```diff
@@ -3480,7 +3481,7 @@
         "mlr3misc",
         "paradox"
       ],
-      "Hash": "9bc0d509207ec242a642de05d40f1f01"
+      "Hash": "705852b917ca86e448bb58258da1f973"
     },
```
RaphaelS1, line 3485: Why is this changed?

sebffischer: this was unintended, I reverted it

PR #739, opened 2023-10-12
fix: set `--without-keep.source` for renv package installations

jemus42:
Addresses https://github.com/rstudio/renv/issues/1713 by setting an option via `.Rprofile` in two separate locations:

a) The project root (where the user (at least I)) tend to start R sessions from
b) `./book/`, where CI starts R sessions from for building the book and calling `renv::restore()`

Please note that for the fix to take effect, all packages need to be installed with the new option, meaning that packages existing in the cache on CI need to be purged.

Locally, you can reinstall all renv packages with `renv::rebuild()`.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,0 +1,4 @@
+# Telling renv to not install with --with-keep.source after all
+# Circumvents https://github.com/rstudio/renv/issues/1713
+# Keep this in each directory where user or CI might call renv::restore() from!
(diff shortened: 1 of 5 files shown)


jemus42:
I should note this _hopefully_ supersedes #736 as it should not be necessary to downgrade renv with this fix.

sebffischer (approved):
lgtm

PR #742, opened 2023-10-22
Update non-sequential_pipelines_and_tuning.qmd

Shitao5:
The addition of a random seed `set.seed(1234)` makes the results of the random search `tnr("random_search")` reproducible and aligns with the current description in the book.

image

image


mb706:
Hi, thanks for the heads-up about the typo. However, it is not actually necessary to set the seed here, as noted [here](https://mlr3book.mlr-org.com/chapters/chapter1/introduction_and_overview.html#reproducibility): the included `_setup.qmd` at the top already sets a seed.

RaphaelS1:
Also just to point out the results are consistent. The number `5` references the above code chunk, plotting then shows you that `4` would be just as good

PR #745, opened 2023-10-24
refactor: solutions of chapter 2

be-marc:


larskotthoff:
The last exercise is a bit weird because the false positive rate is already lower than the false negative rate and changing the threshold doesn't change either. Can we change the exercise? Or at least change the solution to further lower the false positive rate.

be-marc:
That was changed in [fix: exercise 3](https://github.com/mlr-org/mlr3book/pull/745/commits/a57a66061ff9be37af4642f9679c5b8f4f082740).

Issue #747, opened 2023-10-30
Old link to solutions


jemus42:
This link here https://github.com/mlr-org/mlr3book/blob/a901eba0d6ab7f203de28f2599f49506dd959771/book/chapters/chapter1/introduction_and_overview.qmd#L68

Goes to https://mlr3book.mlr-org.com/ but should got to https://mlr3book.mlr-org.com/chapters/appendices/solutions.html

Quarto [redirects](https://quarto.org/docs/websites/website-navigation.html#redirects) can be used for some redirects and we will need to set up some redirects for outdated links anyway.

PR #748, opened 2023-10-30
Add redirect to make old /solutions.html link work again

jemus42:
Addresses #747 by adding a [quarto redirect](https://quarto.org/docs/websites/website-navigation.html#redirects) from `/solutions.html` to the current link.

Note the leading `/` is necessary to have the redirect "start" at the root of the URL, i.e. `mlr3book.mlr-org.com/`

This became necessary as all chapters were restructured into a different folder hierarchy, so we will likely need to add a bunch of aliases for old links to chapters.

### DIFF
--- a/book/chapters/appendices/solutions.qmd
+++ b/book/chapters/appendices/solutions.qmd
@@ -1,6 +1,8 @@
 ---
 params:
     rebuild_cache: true
+aliases:
+  - "/solutions.html"
 ---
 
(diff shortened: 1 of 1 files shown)


jemus42:
Preview confirms it's working as intended - https://748-merge--mlr3book.netlify.app/solutions.html redirects to the correct chapter in the appendix.

PR #753, opened 2023-10-31
fix openml cache paths

sebffischer:
When the folder structure of the book was changed, not only did the hash-keys for the openml cache become unvalid (which I already fixed), also the working directory during rendering changed. Because of this, the openml caching mechanism failed.

We also refactor the solutions, so the ls-benchmarking is its own file, which we can then use to calculate the hash-key for the openml cache. This avoids cache invalidation when working on solutions about chapteters other than the ls-benchmarking chapter.


RaphaelS1:
@sebffischer please could you render locally and screenshot just to show the `include` works as expected and the page and relevant exercises work as intended?

sebffischer:
@RaphaelS1 I did not render it locally, but the netlify preview looks fine. You can check it yourself here: https://753-merge--mlr3book.netlify.app

RaphaelS1 (approved):

PR #758, opened 2023-11-03
adds bibtex citation for all chapters

RaphaelS1:
@jemus42 or @sebffischer would you mind quickly reviewing? Can probably just wait for netlify preview to finish and then look at that


jemus42:
- Added  `here::here()` rather than relative file paths, e.g. `here::here("book/common/chap_auths.csv")` so testing this locally doesn't require `setwd()`
- Added code formatting around the bibtex entry so it looks nicer on the website

These should have been review comments rather than rudely pushing into other people's PRs sorry 🙈

jemus42:
Anyway it looks like this now:

Screenshot 2023-11-03 at 10 35 14

jemus42 (approved):

RaphaelS1:
There's now a '\@' visible, can you remove extra slash?

jemus42:
Oh, whoopsie, fixed it.
Screenshot 2023-11-03 at 13 19 00

jemus42:
These chapter citations use the "old" style link, i.e. `mlr3book.mlr-org.com/<chapter_name>.html` rather than the format we currently have, so I'll set up aliases for these in another PR. 

We talked about needing to do that anyway, and I guess there's no point in waiting.

PR #764, opened 2023-11-06
Actually fix index bibtex title display

jemus42:
The first page displays the bibtex citation which should look like this:

```
@book{Bischl2024
    title = {Applied Machine Learning Using mlr3 in R},
    editor = {Bernd Bischl and Raphael Sonabend and Lars Kotthoff and Michel Lang},
    url = {https://mlr3book.mlr-org.com},
    year = {2024},
    isbn = {9781032507545},
    publisher = {CRC Press}
}
```

but due to an oversight an earlier version used

```
title = {{< meta book.title >}},
```

which was wrongly rendered without the `{ }` braces around the title field, since `{{ }}` are quarto's shortcode delimiters.

I thought I had fixed this in an earlier PR, but I failed to notice that `{{{ }}}` is quarto's way of [escaping shortcodes](https://quarto.org/docs/extensions/shortcodes.html#escaping).
I also failed to noticed that further down, the same content is repeated for the PDF output, where the original formatting error persisted.

I have now opted to not try to use quarto's metadata shortcodes and just plainly inserted the full title of the book where needed, as I do not think it's worth it to spend more time trying to outsmart any shortcode escaping behavior given that the title of the book is not going to change anytime soon 😅 

- [x] Actually triple double check both HTML and PDF version look correct now with these changes

### DIFF
--- a/book/index.qmd
+++ b/book/index.qmd
@@ -280,7 +280,7 @@ Bischl, B., Sonabend, R., Kotthoff, L., & Lang, M. (Eds.). (2024).
 "{{< meta book.title >}}". CRC Press. https://mlr3book.mlr-org.com
 
 @book{Bischl2024
-    title = {Applied Machine Learning Using mlr3 in R}
+    title = {Applied Machine Learning Using mlr3 in R},
     editor = {Bernd Bischl and Raphael Sonabend and Lars Kotthoff and Michel Lang},
     url = {https://mlr3book.mlr-org.com},
     year = {2024},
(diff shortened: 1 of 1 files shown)


jemus42:
There we go, HTML looks fine:

Screenshot 2023-11-08 at 13 20 33

And PDF also looks fine:

Screenshot 2023-11-08 at 13 21 11

RaphaelS1:
I think you need curly brackets around {m}lr3 and {R} to preserve capitalisation?

jemus42:
Oh right, fixed it - should have just leaned on `citeas()` in the first place.

RaphaelS1 (approved):

PR #765, opened 2023-11-06
separate render to pdf and html

sebffischer:


jemus42:
Yep, which made me think about the rendering strategy in general 😬

sebffischer:
I would probably just go for 


```
quarto render book/ --to html
```

as the first step and then 


```
quarto render book/ --to pdf
```

(without the --cache-refresh)

I am not sure whether there are many cases where html or pdf fail without the other format failing as well (unless it is random of course)

sebffischer:
once everything is computed and cached, the second render should be pretty fast I guess, no?

jemus42:
>  once everything is computed and cached, the second render should be pretty fast I guess, no?

Normally, yes.
There are a few things that happen conditionally based on whether the output format is HTML or PDF but I can't think of any major point of weakness in that regard (then again, that would be a reason to test it rather than assuming it's fine)

But yeah I think rendering sequentially in the same workflow but separate steps is simplest and should be fine.

PR #794, opened 2023-11-07
fix(ci): include mermaid diagrams as png and svg (technical)

sebffischer:
Instead of rendering the diagrams using mermaid, they are included as as images.
Note that their look changed slightly, but otherwise they are the same.

For a while, we have suffered from random timeouts when rendering the book. The log-output from GHA during the "Render book" stage just ended with:

|.............................................| 100% output file: advanced_technical_aspects_of_mlr3.knit.md

Error: The operation was canceled.

after hitting the maximum runtime of 6 hours.
(the success rate was around 50/50).

When rendering with the --execute-debug flag
more log-output was given.
The log output at the end of rendering the technical chapter (pdf) was:

 |
  |.............................................| 100%

output file: advanced_technical_aspects_of_mlr3.knit.md

[knitr engine]: writing results
[knitr engine]: exiting
Error: The operation was canceled.

for the other chapters when rendering to pdf, the output was

  |............................................| 100%

output file: preprocessing.knit.md

[knitr engine]: writing results
[knitr engine]: exiting
[knitr engine]: postprocess
[knitr engine]: writing results
[knitr engine]: exiting

--> something with the postprocessing went wrong and the bug was identified to be in the technical chapter

The problem was **NOT** the large-scale benchmarking chapter

https://github.com/quarto-dev/quarto-actions/issues/45


book/chapters/chapter10/Figures/cpu_utilization_1.png
jemus42, line 1: I'm comparing this against Figure 10.2 in the current HTML version and I'm a bit confused by the color difference (the "idle" bars are black here, but dark gray in the book"). I don't think that's a big issue but I'm just wondering how that happened?

.github/workflows/book-weekly.yml
```diff
@@ -57,4 +57,4 @@ jobs:
           restore-keys: ${{ runner.os }}-openml-
 
       - name: Render book
-        run: quarto render book/ --cache-refresh
+        run: quarto render book/ --cache-refresh --execute-debug
```
jemus42, line 60: Is this for the current run only or do we want to enable this permanently?
If it causes some noticeable overhead or very verbose output we might want to remove it once everything's settled again I guess?

jemus42 (approved):
Cool, luckily there's not too many mermaid diagrams so yeah, just pre-rendering them seems like a solid approach.

book/chapters/chapter10/Figures/cpu_utilization_1.png
RaphaelS1: Likely because of this https://github.com/mlr-org/mlr3book/blob/main/book/_quarto.yml#L74-L75

RaphaelS1: @sebffischer can you apply this theme to the new ones?

sebffischer: I used the png from the pdf version and then converted this to svg using inkscape, which change the color. I then exported this svg to png to make sure both versions look the same.

It wasn’t so obvious to me how to render mermaid to an svg with the correct height / width ratio and I think the colours look fine, so I left it that way. If you mind the color I would appreciate that someone else changes it, I already spent enough time figuring out what goes wrong

.github/workflows/book-weekly.yml
sebffischer: hmm, good point about the overhead, it did not seem to drastically increase rendering times and it created some very useful debug info

book/chapters/chapter10/Figures/cpu_utilization_1.png
sebffischer: (or tell me how I can do this :D)

.github/workflows/book-weekly.yml
RaphaelS1: @jemus42 are you happy for this to be merged then?

jemus42: Sure - if it ends up being superfluous, we can still disable debug output down the road, I just wanted to check that's not left in on accident :)

jemus42:
Took the liberty of hitting merge on this one in there interest of fixing CI builds on all the solution-PRs such that Bernd etc. can see previews asap 😬

PR #820, opened 2024-05-30
Clarify encoding pipeline

mb706:
Clarify why later factor encoding pipeops don't need to de-select high cardinality factors.


sebffischer:
@mb706 the CI now runs again. should this be merged?

PR #855, opened 2025-01-09
feat(section): add information about mlr3inferr

sebffischer:


book/chapters/chapter3/evaluation_and_benchmarking.qmd
```diff
@@ -285,6 +285,31 @@ print(plt2)
 ```
 
 
+### Confidence Intervals {#sec-resampling-ci}
+
+Confidence intervals (CIs) provide a range of values within which we can be confident that it covers the true generalization error.
```
larskotthoff, line 290: I would delete the first sentence. The next sentence basically says the same, but this one isn't quite accurate.

book/chapters/chapter3/evaluation_and_benchmarking.qmd
```diff
@@ -285,6 +285,31 @@ print(plt2)
...
+Instead of relying solely on a single point estimate, CIs offer a measure of uncertainty around this estimate, allowing us to understand the reliability of our performance estimate.
+While constructing CIs for the generalization error is challenging due to the complex nature of the inference problem, some methods have been shown to work well in practice @kuempelfischer2024ciforge.
```
larskotthoff, line 292: I would add some more context here -- some learners/models can provide these directly (and often those calculations aren't all that complex), but if the learner doesn't support it, we have to do something else. Then describe in a sentence or two what those methods do.

book/chapters/chapter3/evaluation_and_benchmarking.qmd
```diff
@@ -285,6 +285,31 @@ print(plt2)
...
+As opposed to other mlr3 measures, the result is not a scalar value, but a vector containing the point estimate, as well as the lower and upper bounds of the CI for the specified confidence level.
+
+```{r}
+library(mlr3inferr)
+# alpha = 0.05 is also the default
+msr_ci = msr("ci.wald_cv", msr("classif.acc"), alpha = 0.05)
+rr$aggregate(msr_ci)
+```
+
+We can also use `msr("ci")`, which will automatically select the appropriate method based on the `Resampling` object, if an inference method is available for it.
```
larskotthoff, line 307: How do I know what resamplings have inference methods?

book/chapters/chapter3/evaluation_and_benchmarking.qmd
```diff
@@ -576,6 +601,22 @@ plt = plt + ggplot2::scale_fill_manual(values = c("grey30", "grey50", "grey70"))
 print(plt)
 ```
 
+It is also possible to plot confidence intervals by setting the type of plot to `"ci"`.
+Ignoring the multiple testing problem, @fig-benchmark-ci shows that the difference between the random forest and both other learners is statistically significant for the sonar task, whereas no final conclusion can be drawn for the german credit problem.
```
larskotthoff, line 605: Can we not ignore the multiple testing problem? I would show results for a single learner here.

larskotthoff (changes requested):

book/chapters/chapter3/evaluation_and_benchmarking.qmd
```diff
@@ -285,6 +285,31 @@ print(plt2)
 ```
 
 
+### Confidence Intervals {#sec-resampling-ci}
```
sebffischer, line 288:
```suggestion
### Confidence Intervals (+) {#sec-resampling-ci}
```

PR #881, opened 2025-06-16
add warning that validation scores of graphlearner are prefixed by PO id

sebffischer:


book/chapters/chapter15/predsets_valid_inttune.qmd
```diff
@@ -302,6 +302,10 @@ ti = tune(
 )
 ```
 
+::: {.callout-warning}
+When working with a `r GraphLearner`, the names of the internal validation scores are prefixed by the ID of the corresponding `r PipeOp`, so the `select` parameter needs to be set to `"<pipeop id>.<measure id>"`.
```
sebffischer, line 306:
```suggestion
When working with a `GraphLearner`, the names of the internal validation scores are prefixed by the ID of the corresponding `PipeOp`, so the `select` parameter needs to be set to `"<pipeop id>.<measure id>"`.
```

